PR #21, opened 2015-11-16
Race condition during make install-data

nabilbendafi:
*pspmpegbase.h* header is included twice: once as *libpspmpegbaseinclude_HEADERS* and once as *libpspmpegbase_driverinclude_HEADERS*, in `src/mpeg/Makefile`

When Travis_CI runs `make -j 32 install-data`, race condition occurs on that file, and installation fails

Is it really needed to have it defined such way ?


artart78:
Well, no idea why it was defined like this, but maybe there was a pspmpegbase_driver.h which has been removed, and the fix hasn't been done properly. Anyway, it seems to be fine to remove these lines.

nabilbendafi:
Thanks.

PR #25, opened 2018-08-18
Remove the `-mno-crt0` flag from build_prx.mak

### LINKED ISSUE #24: build_prx.mak: -mno-crt0 option doesn’t exist in current psp-gcc
`-mno-crt0` does not exist in current versions of psp-gcc installed by the toolchain.

[`build_prx.mak` is still dependent on this flag](https://github.com/pspdev/pspsdk/blob/300ec5c405d7c41d43c6735c5897aa80c731b7c0/src/base/build_prx.mak#L32), and for plugins whose makefiles inherit from it correctly, this causes building them to fail.

This appears to be a long-standing issue, with no satisfactory solution given: pspdev/psptoolchain#72, pspdev/psptoolchain#78

ticky:
`-mno-crt0` does not exist in current versions of psp-gcc installed by the toolchain, and `-nostartfiles` appears to be equivalent.

fixes #24

### DIFF
--- a/src/base/build_prx.mak
+++ b/src/base/build_prx.mak
@@ -29,7 +29,7 @@ CFLAGS   := $(addprefix -I,$(INCDIR)) $(CFLAGS)
 CXXFLAGS := $(CFLAGS) $(CXXFLAGS)
 ASFLAGS  := $(CFLAGS) $(ASFLAGS)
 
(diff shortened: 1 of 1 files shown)


liclac:
Pinging @artart78

artart78:
I don't have enough free time to test the absence of regressions for every pull request as I'd like to, but this seems safe enough :P

carstene1ns:
Yay!

PR #34, opened 2020-05-08
psp-prxgen: Allow relocations before relocated section

overdrivenpotato:
This commit scans through all sections first, before scanning for relocation sections. Scanning relocations in a separate step allows relocations to appear before the section they are relocating (valid with ELF).

### DIFF
--- a/tools/psp-prxgen.c
+++ b/tools/psp-prxgen.c
@@ -281,7 +281,11 @@ int load_sections(unsigned char *data)
 						load_addr = g_elfsections[i].iAddr;
 					}
 				}
+			}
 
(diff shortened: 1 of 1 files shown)


carstene1ns:
The change seems reasonable. Do you have a testcase that breaks the current behavior?

overdrivenpotato:
Here’s an example of an ELF file linked with LLD instead of psp-ld. This fails in psp-prxgen without the patch, as all the relocations occur at the top.

image

overdrivenpotato:
Is there anything else that I can do to help have this PR merged?

carstene1ns:
I just wanted to test it locally first. Thanks for your contribution! :+1:

PR #41, opened 2020-10-03
Fix doxygen addgroup syntax

sajattack:
This fixes the problem we've been experiencing where the module pages are empty.


wally4000 (approved):
Type correct, approved.

PR #49, opened 2021-01-11
Fix pspDebugGetStackTrace() by using _etext as upper bound for ra

yoanlcq:
Hi,

As in the title, this is a fix for pspDebugGetStackTrace(). It never worked for me in the first place, maybe because it's only supposed to work with ELF executables instead of PRX, or some other reason I don't know.  

Anyway, the main issue was the use of a hardcoded address as the supposed upper bound for the return address (ra) value, which happened to be wrong. I think the correct thing to do is to use the address of `_etext` instead, which should always reliably give the address of the end of the `.text` section.

As far as my tests go, the function works with this fix. It seems to stop before adding `main` to the list, but as far as I'm concerned, it's better than `it doesn't work`! :)
Thanks!

### DIFF
--- a/src/debug/callstack.c
+++ b/src/debug/callstack.c
@@ -128,6 +128,10 @@ typedef int Bool;
 #define FALSE 0
 #endif
 
+// Defined by the linker, denotes the start and end of the .text section
+extern unsigned int _ftext;
+extern unsigned int _etext;
+
 extern unsigned int *pspGetReturnAddress();
 extern unsigned int *pspGetStackPointer();
 extern int main();
@@ -135,7 +139,7 @@ extern int main();
 int pspDebugGetStackTrace(unsigned int *results, int max)
 {
 	unsigned int *ra;
-	unsigned int *ra_limit;
+	unsigned int *ra_limit = &_etext;
 	unsigned int *sp;
 	unsigned int inst;
 	unsigned int mainCall;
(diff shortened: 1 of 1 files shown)


src/debug/callstack.c
```diff
@@ -128,14 +128,18 @@ typedef int Bool;
...
+
 extern unsigned int *pspGetReturnAddress();
 extern unsigned int *pspGetStackPointer();
 extern int main();
 
 int pspDebugGetStackTrace(unsigned int *results, int max)
 {
 	unsigned int *ra;
-	unsigned int *ra_limit;
+	unsigned int *ra_limit = &_etext;
```
carstene1ns, line 142: This is actually not needed as it is set in the loop.

carstene1ns (commented):
It looks good to me. Unfortunately I never worked with the stacktrace.

I guess a simple testcase would be nice (will see if i can come up with one).

src/debug/callstack.c
yoanlcq: I thought so, too, but GCC gave a warning about the variable being possibly used while uninitialized.  
Indeed it is possible to reach line 244 without going though the if() at line 162. :)

carstene1ns: You are right about that. :facepalm:

MotoLegacy:
@carstene1ns seems this PR has been held hostage for a couple years now. Would be nice if the review was edited to be marked as approved given the context in the comments so the ball could get rolling again?

sharkwouter (approved):
I am not really able to test this, but since this affects a function which reportedly does not work and the code has already been reviewed before, I'm just going to merge it.

PR #54, opened 2021-04-05
pspctrl.h: Add sceCtrlLatch API documentation

Felix-Dev:
This PR aims to add documentation for the *Latch APIs in the controller service library. The documentation has been [taken from uOFW](https://github.com/uofw/uofw/blob/master/include/ctrl.h) and should be correct.

The doxygen output can be found [here](https://uofw.github.io/uofw/group__Controller.html#ga3a3ffd1ee4d1d6e93dfb95b5df8dfaaf).

@sajattack FYI.

### DIFF
--- a/src/ctrl/pspctrl.h
+++ b/src/ctrl/pspctrl.h
@@ -101,10 +101,45 @@ typedef struct SceCtrlData {
 	unsigned char 	Rsrv[6];
 } SceCtrlData;
 
+/** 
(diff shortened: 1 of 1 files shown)


sajattack:
I'm not certain if our doxygen is configured identically to yours but if it's not I'll fix it after the fact.

sajattack (approved):

Issue #56, opened 2021-10-09
Printf function isn't working


fjtrujy:
During the migration and upgrade of newlib I have faced that `printf` function is not working anymore in the current toolchain.
(@wally4000 double-checked it.)

I have been investigating for some days till I found why it wasn't working.

On the `crt` files we have `__psp_libc_init` as `weak reference`
https://github.com/pspdev/pspsdk/blob/master/src/startup/crt0.c#L41
```c
/* Allow newlib/psplibc to provide an init hook to be called before main */
extern void __psp_libc_init(int argc, char *argv[]) __attribute__((weak));
```

Then in the `_start` method (entrance point of every single execution), we have:
https://github.com/pspdev/pspsdk/blob/master/src/startup/crt0.c#L78
```c
	/* Call libc initialization hook */
	if(__psp_libc_init != NULL)
		__psp_libc_init(argc, argv);
```

Till this point everything looks fine, I think that the developer who implement this, was thinking:
 _"if user links the binary with `libpsplibc.a` then it will call the `__psp_libc_init` function"_

Then on the `build.mak` that we usually use for compilation we have:
https://github.com/pspdev/pspsdk/blob/master/src/base/build.mak#L72
```make
ifeq ($(USE_PSPSDK_LIBC),1)
# Use the pspsdk libc
PSPSDK_LIBC_LIB = -lpsplibc
CFLAGS := -I$(PSPSDK)/include/libc $(CFLAGS)
else
```

Cool! Everything still looks fine, where is the problem then?
The problem is that the `__psp_libc_init` definition in the `libpsplibc.a` is as well being `weak`:
https://github.com/pspdev/pspsdk/blob/master/src/libc/init.c#L20
```c
__attribute__((weak, constructor))
void _pspsdk_libc_init()
{
    _pspsdk_alloc_init();
    _pspsdk_stdio_init();
    _pspsdk_stdlib_init();
}
```

So basically all the references of `__psp_libc_init` in the binary are weak, with the consequences of `__psp_libc_init` never being executed even if the user adds the `libpsplibc.a` during the linking process.

In my opinion, there are 2 solutions:

1.  To remove the weak refecence in the crt files
2.  To remove the weak reference in the libpsplibc.a

I honestly prefer the first one, and it is how I have done in the toolchain that I'm preparing.

Thanks

liclac:
This is some impressive detective work!
Doing that makes sense to me. 👍

carstene1ns:
[...]

PR #58, opened 2021-11-13
Update PSPSDK and make it more POSIX

fjtrujy:
# Description

First of all sorry for this huge PR, but it wasn't possible to do it shorter.

The original scope of this PR was to perform the needed changes to implement in `libpsplibc` the required `newlib` callbacks for creating a fully POSIX environment, however in the path, other changes have been needed.
 
Most important items:
- Add some missing headers and stubs to some user and kernel libs (LightWidth Mutex).
- Migrate code from the newlib patch to the `libcglue` (needed by decouple `psptoolchain` from `pspsdk`)
- Improve the logic for making thread-safe all the I/O operations and malloc/free
- Improve the logic for reserving the available `heap memory`, now by default uses everything except 512KB reserved just in case is needed for other libraries.
- Improve base makefiles, making them more simple because now `gcc` includes the standar libraries (https://github.com/pspdev/gcc/blob/133f2d65bee538e1141a9701f7f80dd4b21649ce/gcc/config/mips/psp.h#L23-L30)
- Fix an issue with `__psp_libc_init` not being called
- Add `cmake` files to base folder
- Automate generation of `doxygen`
- Compile in the CI/CD all the examples generating the proper artifacts.
- Create `pspsdk` layer 
- Include in the CI/CD compilation using the `psptoolchain` docker layer.
- Other minor issues

## POSSIBLE BREAKING CHANGES
1. Some fields from `SceIoStat` have been renamed because entered in conflicts with `newlib`
```
st_ctime -> sce_st_ctime
st_atime -> sce_st_atime
st_mtime -> sce_st_mtime
```

2. Heap allocation logic has been changed. 
Previously we just have a macro called `PSP_HEAP_SIZE_MAX` for using the whole `heap memory`, now we have a  macro called `PSP_HEAP_THRESHOLD_SIZE_KB`, where you can define the `threshold you leave available`. The default value is **`512KB`**
So, `PSP_HEAP_THRESHOLD_SIZE_KB(0)` is exactly the same that before was `PSP_HEAP_SIZE_MAX()`

3. Removed macro `PSPSDK_LIBC_LIB` and `USE_KERNEL_LIBC`. 
By default always link with `libc.a + libcglue.a` & using standard libraries, however, the option `USE_KERNEL_LIBS` is still available and it will always use the `Internal Kernel Libc`

4. Some other unexpected issues.

### MORE INFO
This PR is part of a collection of PRs, where the whole `pspdev` environment for development is being improved, with the main objectives of getting a more POSIX & Standard toolchain and integrating a CI/CD workflow to increase testability and reliability.

Others PRs related:
https://github.com/pspdev/psp-pacman/pull/15
https://github.com/pspdev/psptoolchain-extra/pull/1
https://github.com/pspdev/psptoolchain-allegrex/pull/1
https://github.com/pspdev/psptoolchain/pull/160

Here you have more info about the new repo structures, artifacts, and docker layers generated:

layers

Thanks


sajattack (changes requested):
Leaving my comment from discord here as well so it's not forgotten:

What you're calling psplibc here isn't a full libc and is very different to the previous psplibc (I think). I would like it renamed. Something like libcglue or libposixinterface or whatever we can come up with.

src/libcglue/errno.c
```diff
@@ -3,18 +3,24 @@
  * -----------------------------------------------------------------------
  * Licensed under the BSD license, see LICENSE in PSPSDK root for details.
  *
- * unistd.h
+ * init.c - The global init/deinit code for our crt0.
```
carstene1ns, line 6: errno.c

configure.ac
```diff
@@ -70,7 +70,7 @@ AC_SUBST(PSPSDK_INCLUDEDIR)
 AC_SUBST(PSPSDK_LIBDIR)
 
 # CFLAGS and CXXFLAGS used to build pspsdk libraries.
-PSPSDK_CFLAGS="$CFLAGS -G0 -Wall"
+PSPSDK_CFLAGS="$CFLAGS -G0 -O3 -Wall"
```
carstene1ns, line 73: please remove, so the user can have their own (e.g. `-O0` for debugging), autotools will automatically set `-O2` when nothing is set.

src/libcglue/cwd.c
```diff
@@ -0,0 +1,192 @@
+/*
+ * PSP Software Development Kit - https://github.com/pspdev
+ * -----------------------------------------------------------------------
+ * Licensed under the BSD license, see LICENSE in PSPSDK root for details.
+ *
+ * init.c - The global init/deinit code for our crt0.
```
carstene1ns, line 6: cwd.c

src/libcglue/timezone.c
```diff
@@ -0,0 +1,34 @@
+/*
+ * PSP Software Development Kit - https://github.com/pspdev
+ * -----------------------------------------------------------------------
+ * Licensed under the BSD license, see LICENSE in PSPSDK root for details.
+ *
+ * timezon.c - Set proper timezone needed by newlib.
```
carstene1ns, line 6: timezone.c

src/samples/Makefile.am
```diff
@@ -84,12 +84,11 @@ SAMPLES = \
...
 	exceptions
 
-# FIXME: This workaround is actually needed to get the stage1 installation
-# without samples working. Once a better way to do header installation for
-# newlib is figured out, this can be rewritten.
-
 all:
 
+dist-hook:
+	rm -rf `find $(distdir) -name .svn -o -name CVS`
```
carstene1ns, line 90: There should be no traces of subversion or cvs around anymore. Why do you revert (https://github.com/pspdev/pspsdk/commit/d88103a9be5bde6599a279cbdfe1b080fb87dc09#diff-3368d769c27792b586069451459f50fb33504e5b99bcfd17e841dd7a9e3128cd) to an older version and stripping my comment?

Makefile.am
```diff
@@ -4,6 +4,6 @@ SUBDIRS = src tools
 
 MOSTLYCLEANFILES = $(DX_CLEANFILES)
 
-EXTRA_DIST = LICENSE VERSION README.md Doxyfile doc
+EXTRA_DIST = LICENSE VERSION README.md Doxyfile doc/html
```
carstene1ns, line 7: This likely breaks `make distcheck`

src/startup/Makefile.am
```diff
@@ -10,21 +10,13 @@ CCASFLAGS = $(CFLAGS)
 
 # Borrowed this trick from newlib: create a dummy library to generate build rules.
 noinst_LIBRARIES = lib.a
-lib_a_SOURCES = dummy.c
-
-# FIXME: This workaround is actually needed to get the stage1 installation
-# working. Once a better way to do crt0 installation for newlib is figured out,
-# this can be rewritten.
+lib_a_SOURCES = crt0.c crt0_prx.c prxexports.c
```
carstene1ns, line 13: Does this work?

carstene1ns (commented):
some nits.

src/startup/Makefile.am
fjtrujy: Yes, I tried with several applications, ELF and PRXs both working

src/libcglue/errno.c
fjtrujy: Thanks

carstene1ns:
Can you please explain why the weak reference to `__psp_libc_init` does not work like before and we need to resort to another macro when wanting to disable it?

src/startup/Makefile.am
carstene1ns: So no need to feed the crt0.o into a first stage installation anymore?

fjtrujy: Nop. I recommend you to take a look to the graph attached here:

The toolchain is composed by:
1. Binutils 
2. Gdb
3. Gcc phase 1
4. Newlib
5. Gcc phase 2

This was one of the most important things why I wanted to do that, fully separate pspsdk from psptoolchain

fjtrujy:
> Can you please explain why the weak reference to `__psp_libc_init` does not work like before and we need to resort to another macro when wanting to disable it?

So there is more than one topic here:
First this one, it wasn't working properly
https://github.com/pspdev/pspsdk/issues/56

Then secondly, now by default, we always link using `newlib` unless we decide to use `USE_KERNEL_LIBS`
So by default the `__psp_libc_init` must be called, if for whatever reason, you don't want to call this init, you still can override the function because `__psp_libc_init` is defined as weak in `src/libcglue/init.c`.

The reason why I created that macro is like a "shortcut" for users that fully want to disable the `psp_libc_init` (only good reason is if you aren't going to use anything about `newlib` and you want to have a super tiny binary)

fjtrujy:
> Leaving my comment from discord here as well so it's not forgotten:
> 
> What you're calling psplibc here isn't a full libc and is very different to the previous psplibc (I think). I would like it renamed. Something like libcglue or libposixinterface or whatever we can come up with.

We have finally renamed to `libcglue`, thanks for the suggestion

src/samples/Makefile.am
fjtrujy: I think it was probably a mistake that I did. Removing thing that.
The comment is not anymore needed, because I have fixed that you said there 😄

Makefile.am
fjtrujy: Thanks, fixing it!

fjtrujy:
Hello @carstene1ns,
I have renamed `__psp_libc_init` to `__libcglue_init`, maybe now is clearer to understand what does it means.

Thanks

src/libcglue/glue.c
```diff
@@ -0,0 +1,865 @@
...
+{
+	// TODO: Implement proper functionality
+    return 0;
+}
+#endif
+
+#ifdef F_fchmod
+int fchmod(int filedes, mode_t mode)
+{
+	// TODO: Implement proper functionality
```
sajattack, line 804: TODOs?

sajattack (approved):

PR #60, opened 2021-11-17
Add CreatePBP.cmake and add some variables relied on by software

sharkwouter:
This is what I need to be able to use cmake for my projects. My projects will still require some minor changes, since my old cmake files were very specific for the old toolchain, but it should mostly be good.

I really need ``__PSP__`` and ``PLATFORM_PSP`` to be set in the same way as in the previous toolchain.

I do miss the psp-cmake script as well, but if we want to re-add that, I'll do so in a separate PR.


fjtrujy (approved):
🚀

PR #61, opened 2021-11-23
Fix gettimeofday & other small improvements

fjtrujy:
# Description

This PR fixes a wrong implementation of `int _gettimeofday(struct timeval *tp, struct timezone *tzp)` glue function.

`Newlib` expects to have in the `struct timeval` the `tv_sec` parameter as 64bits item, however `PSP OS`, when calling `sceKernelLibcGettimeofday`, it expects `tv_sec` to be 32bits. Similar issue happens with `select` function for `inet`.

Other minor improvements have been done.


wally4000 (approved):
Looks sane :)

PR #63, opened 2021-11-25
Improve Exit/Terminate process

fjtrujy:
This PR just basically improves the exit process, making sure that some objects are `deinit` properly before killing the app.

Thanks


sharkwouter (approved):
I just did at test with this and it works well.

carstene1ns (approved):
looks good to me.

PR #65, opened 2021-12-03
Remove specific condition for heap and PRX files

fjtrujy:
# Description

There was a specific condition in the `PRX` files for having by default `64 KB of HEAP`. I created this condition because I thought was mandatory, however, after some testing, I faced that it wasn't needed.

The test was done together with @davidgf

Thanks

### DIFF
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -41,13 +41,11 @@
 
 #include "fdman.h"
 
-#define DEFAULT_PRX_HEAP_SIZE_KB 64
 #define DEFAULT_HEAP_THRESHOLD_SIZE_KB 512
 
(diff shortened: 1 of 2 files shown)


davidgfnet:
LGTM!
As far as I know kernel PRXs won't use this at all since they do not use libc
For user space PRX it is fine to take most/all of the heap, I don't think we have many or any user PRX modules around. Perhaps some plugin might be sideload with a regular App?
Anyway I'd say we merge this and get the previous behaviour, which is more useful to developers.

davidgfnet (approved):

fjtrujy:
@sharkwouter ⬆️

wally4000 (approved):

PR #66, opened 2021-12-03
Move inline stubs to a separated asm file

davidgfnet:
Defining asm labels in inline assembly is not a good idea, according to
GCC docs it can result in labels being deleted (optimized away) or even
duplicated (see https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html).

This affects LTO linking, since LTO removes the four symbols and results
in undefined references. Using an ASM file makes sure that symbols exist
with the correct attributes.


fjtrujy:
Thanks @davidgfnet ,
`LTO` feature is quite interesting, it could give up some performance improvement 🎉

fjtrujy (approved):

PR #67, opened 2021-12-04
Add pspsdk to the standard libs

fjtrujy:


wally4000 (approved):
Defo a requirement

John-K:
How would one compile anything without including libpspsdk?

PR #73, opened 2021-12-09
Add some more POSIX API calls

davidgfnet:
This adds some stuff used by libconfuse, but potentially other programs that attempt to find the user's HOME dir by parsing the password structure.
No PSP homebrew should use these functions, but there are always some off-the-shelf libs that attempt to use this as a POSIX platform :)

### DIFF
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -22,6 +22,7 @@
 #include <dirent.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <pwd.h>
 
 #include <sys/time.h>
 #include <sys/timeb.h>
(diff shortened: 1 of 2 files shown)


src/libcglue/glue.c
```diff
@@ -926,4 +927,33 @@ int getentropy(void *buffer, size_t length) {
 
 	return 0;
 }
-#endif
\ No newline at end of file
+#endif
+
+#ifdef F_getuid
+uid_t getuid(void) {
+	/* Not sure if returning root is a good idea */
```
sajattack, line 934: 1000 might be a better default. It doesn't really matter but some programs may run differently if they think they're root.

davidgfnet: Yes indeed! I was not sure what the best behaviour was to be honest. (I know some programs refuse to run as root, others assume certain permissions if UID is zero, etc). Lemme change it and we can always change it back if we feel like it :) Thanks!

fjtrujy (approved):

fjtrujy:
Thanks @davidgfnet ,
Every time a POSIXer environment!
Merging it!

PR #75, opened 2021-12-15
Don't make psp_artifacts directory

sharkwouter:
This makes it much easier to use psplink, since you'll no longer need to copy the prx into the build directory upon each build.


sharkwouter:
I've added the option BUILD_PRX which overwrites if a prx should be generated by the create_pbp_file macro.

fjtrujy (approved):

fjtrujy:
Well done @sharkwouter,
Now I think it is cleaner and more flexible

PR #76, opened 2021-12-23
Set CMAKE_INSTALL_PREFIX correctly

sharkwouter:
CMAKE_TARGET_INSTALL_PREFIX isn't actually documented anywhere on the cmake website, so I guess that was a mistake. This change does not seem to cause warnings while building applications.


sharkwouter:
To clarify, at the moment the CMAKE_INSTALL_PREFIX will need to be set manually, otherwise cmake will try to install in /urs/local.

fjtrujy (approved):

fjtrujy:
> To clarify, at the moment the CMAKE_INSTALL_PREFIX will need to be set manually, otherwise cmake will try to install in /urs/local.

Good catch! It was my fault, actually, we're suffering the same error in `PS2`.

Know @sharkwouter there are several places in `psp-packages` where remove this.

sharkwouter:
I think the ones in psp-packages set it to the build directory, so that should probably stay.

PR #85, opened 2022-03-15
Use psp-pkg-config for FindPkgConfig in CMake

sharkwouter:
This make it so that when using the following code:

```
include(FindPkgConfig)
pkg_search_module(SDL2 REQUIRED sdl2)
```

The variables ``${SDL2_INCLUDE_DIRS}`` and ``${SDL2_LIBRARIES}`` are set to the expected values within CMake.


fjtrujy (approved):
Thanks @sharkwouter

glebm:
The cache part is in fact needed, otherwise it only works for the first call to `pkg_search_module`

sharkwouter:
Oh really? That would explain an issue I've seen happen with pkgconfig in cmake. @glebm could you make a PR or should I?

PR #86, opened 2022-03-17
some tiny compatibility fixes

rofl0r:
this fixes build with last year's newlib 1.20.0 based toolchain.

### DIFF
--- a/src/libcglue/fdman.h
+++ b/src/libcglue/fdman.h
@@ -31,11 +31,11 @@ typedef enum {
 } __fdman_fd_types;
 
 typedef struct {
-	u_int32_t descriptor;
-	u_int32_t flags;
-	u_int32_t ref_count;
+	uint32_t descriptor;
+	uint32_t flags;
+	uint32_t ref_count;
 	char *filename;
-	u_int8_t type;
+	uint8_t type;
 } __descriptormap_type;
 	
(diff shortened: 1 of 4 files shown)


rofl0r:
note that the last commit breaks compilation with the latest newlib 4.0.1 allegrex toolchain.

probably should add some `#ifndef` to fix that.

fjtrujy:
I'm confused with the description of that PR, what are you trying to solve?
At the beginning looks it was fixing something in newlib 1.20 and then in the comments you mentioned about newlib 4.0.1 

Could you explain better what is the issue you are suffering and the environment?

Thanks

fjtrujy:
Just for adding more information, the current status of the PSPSDK is not though to be compatible with legacy version of newlib.
Actually `pspsdk` and `newlib fork` we assure they are working together just with current status of both repos.

We can't be supporting retro-compatibilities, it doesn't make any sense to put effort in that kind of situations, there aren't benefits at all....

rofl0r:
> We can't be supporting retro-compatibilities

i understand. i opened this mainly as a documentation for people using the older toolchain.
you might want to consider cherry-picking https://github.com/pspdev/pspsdk/pull/86/commits/181c7b91b7caf1e2ef1d43d9f93bf3336597a2fb so the stdint types are used everywhere for consistency. else i'll just close this.

rofl0r:
closed by #94

PR #88, opened 2022-03-23
improve psp-fixup-imports

rofl0r:


fjtrujy:
Once again, could you explain what you are trying to solve here?

I don't have anything against the changes but no one suffered issues with that tool in the latest years.

Please expose your problems and what does the PR to fix it

rofl0r:
the first 2 commits just make the code more readable by essentially replacing

```c
if (foo != NULL) {
   // indented block
} else return error;
```

with

```c
if(foo == NULL) return error;
// non-indented block
```

the 3rd commit allows copy/pasting the hex offsets printed in verbose mode into readelf | less whose search function is case-sensitive, by using lowercase hex - this is imo the only commit that could be considered offending.

the 4th commit improves the verbose output by adding symbol names for the stubs that are patched. this at least to me appears to be a real value addition.

the 5th commit fixes the actual issue i encountered - when using psp-fixup-imports on the elf's created by pspgl tests/ Makefile, it throws the error "could not fixup imports", even though the binaries work just fine when ignoring the error instead of exiting with 1.
it seems under some rare circumstances the check raising the error is wrong, and i'm not the only one here: https://psp.scenebeta.com/node/51417

KorigamiK:
Does this require any more changes?

rofl0r:
@KorigamiK not from my end, the PR is perfect as is.

sajattack:
> the first 2 commits just make the code more readable by essentially replacing
> 
> ```c
> if (foo != NULL) {
>    // indented block
> } else return error;
> ```
> 
> with
> 
> ```c
> if(foo == NULL) return error;
> // non-indented block
> ```
Minor point but the all-on-one-line syntax makes it more difficult to put a breakpoint on the true case of the  if statement in a debugger

sharkwouter (approved):

PR #90, opened 2022-03-25
Adding support for `std:thread`

fjtrujy:
## Description
This PR is part of a collection for the std::tread support in the PSP Toolchain

This PR follows a similar approach as `newlib` in the implementation plan.

As a requirement we **MUST** have `psptoolchain` layer totally agnostic to `pspsdk`, we should be able to compile `psptoolchain` without knowing anything about `pspsdk` yet.
So in order to achieve this with `pthread-embedded` what I have done is to remove from the `pthread-embedded` repository all the `psp` specific code, and it is under the `pspsdk/src/libpthreadglue`

So basically :
- `pthread-embbeded` generates `libpthread.a`, which call some functions that now are within `pspsdk/src/libpthreadglue`
- `pspsdk/src/libpthreadglue` generates `libpthreadglue.a` which uses `PSP OS` functions to deal with threading.

Why `pthread-embedded` is going to be in `psptoolchain` instead of being as `package` and allocate in `psp-packages` as currently, it is?

As you probably know, in the toolchain we have 2 phases for `GCC`. The first phase allow us to use `C` code, and the second phase is for providing `CPP` support.
In order to make `phase 2` to work, we used to require to have `newlib` support, now If we want to support `std::thread` that requires also to have `pthread-embbeded` in advance.
This is why `pthread-embbeded` is going to be in the toolchain and before `gcc phase 2`.

The majority of the changes that you see have been basically taken from the `pthread-embedded` for the specific `psp` platform code.

⚠️ IMPORTANT:
Now `libpthread.a` and `libpthreadglue.a` is going to be part of the `standard libraries`, it means that you won't even need to add anything for using them, is going to be out of the box.
It means, that binary size will increase automatically around 15KB, however I'm offering a way to disable the `pthread` support if you aren't going to use it.
You just need to put in your code the next macro (usually in the `main.c` file):
```c
PSP_DISABLE_AUTOSTART_PTHREAD();
```

That's actually needed because during the initialization process we need to call explicitly the function `pthread_init()`, otherwise `pthread` and obviously `std::thread` won't work.
My intention in the future is to improve the initialization of the `pthread` library to make the initialization automatically, and transparent to the user.

### DIFF
--- a/src/libcglue/sleep.c
+++ b/src/libcglue/sleep.c
@@ -19,6 +19,9 @@
 /* Fuctions from errno.c */
 int __set_errno(int code);
 
+/* Fuctions from errno.c */
+int __set_errno(int code);
+
 #ifdef F_nanosleep
 /* note: we don't use rem as we have no signals */
 int nanosleep(const struct timespec *req, struct timespec *rem)
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,10 @@ RUN cd /src && \
     make -j $(getconf _NPROCESSORS_ONLN) && \
     make -j $(getconf _NPROCESSORS_ONLN) install
 
-## gcc needs to include libcglue libpsputility libpsprtc libpspnet_inet libpspnet_resolver libpspmodinfo libpspuser libpspkernel
+## gcc needs to include libcglue libpthreadglue libpsputility libpsprtc libpspnet_inet libpspnet_resolver libpspmodinfo libpspuser libpspkernel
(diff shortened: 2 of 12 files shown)


src/libcglue/sleep.c
```diff
@@ -9,6 +9,7 @@
  *
  */
 
+#include <stdio.h>
```
sajattack, line 12: looks like this is included twice.

fjtrujy: Fixed!

sajattack (approved):

PR #93, opened 2022-03-30
Adding `clock_getres`, `clock_gettime` and `clock_settime` to `libcglue`

fjtrujy:
## Description

This PR basically adds some missing functions in newlib to the `libcglue` library to make it more POSIX.


wally4000 (approved):
Passes CI and looks sane.

PR #95, opened 2022-04-06
Remove 512 width/height limitation on the sceGuTexImage

fjtrujy:
## Description

Following the documentation, in sceGuTexImage function, the maximum width/height looks to be `512`.
https://github.com/pspdev/pspsdk/blob/master/src/gu/pspgu.h#L1185-L1199 

However there are too many games/apps where the minimum texture resolution is 640x480  😢 , I was wondering if that 512 is a real hardware limitation or not...

Taking a look at the implementation of sceGuTexImage I see that we have a condition of 2^9 = 512 for texture width/height.
https://github.com/pspdev/pspsdk/blob/master/src/gu/sceGuTexImage.c#L15

I have removed that condition to check if I could load for instance a texture of 1024x1024, and it worked, I have tested it in PPSSPP and real hardware.

Anyway, the current `pspsdk` behavior is `unexpected` when a user tried a bigger texture than `512`, with this PR at least we are sending to the GPU the proper value, then its up to the hardware to work or doesn't...

Thanks

### DIFF
--- a/src/gu/sceGuTexImage.c
+++ b/src/gu/sceGuTexImage.c
@@ -12,16 +12,9 @@ static int tbpcmd_tbl[8] = { 0xa0, 0xa1, 0xa2, 0xa3, 0xa4, 0xa5, 0xa6, 0xa7 };		
 static int tbwcmd_tbl[8] = { 0xa8, 0xa9, 0xaa, 0xab, 0xac, 0xad, 0xae, 0xaf };		// 0x30A38
 static int tsizecmd_tbl[8] = { 0xb8, 0xb9, 0xba, 0xbb, 0xbc, 0xbd, 0xbe, 0xbf };	// 0x30A58
 
-int getExp(int val)
+static inline int getExp(int val)
 {
-	unsigned int i;
-	asm("clz %0, %1\n":"=r"(i):"r"(val&0x3FF));
-	return 31-i;
-/*
-	unsigned int i;
-	for (i = 9; (i > 0) && !((val >> i) & 1); --i);
-	return i;
-*/
+	return 31 - __builtin_clz(val);
 }
 
 void sceGuTexImage(int mipmap, int width, int height, int tbw, const void* tbp)


artart78:
I guess it would be nice to check on hardware what's the maximum usable value? The field seems to go to up to 16 bits.

fjtrujy:
> I guess it would be nice to check on hardware what's the maximum usable value? The field seems to go to up to 8 bits but 2^31 is far too big already, I wonder where's the actual limit.

Based on my actual tests is actually 2037 which is 10 bits.....

src/gu/sceGuTexImage.c
```diff
@@ -12,16 +12,9 @@ static int tbpcmd_tbl[8] = { 0xa0, 0xa1, 0xa2, 0xa3, 0xa4, 0xa5, 0xa6, 0xa7 };
 static int tbwcmd_tbl[8] = { 0xa8, 0xa9, 0xaa, 0xab, 0xac, 0xad, 0xae, 0xaf };		// 0x30A38
 static int tsizecmd_tbl[8] = { 0xb8, 0xb9, 0xba, 0xbb, 0xbc, 0xbd, 0xbe, 0xbf };	// 0x30A58
 
-int getExp(int val)
+static inline int getExp(int val)
 {
-	unsigned int i;
-	asm("clz %0, %1\n":"=r"(i):"r"(val&0x3FF));
```
sharkwouter, line 18: Wouldn't it be better to change this check rather than removing it? There is still a hardware limit, it's just bigger than we thought.

sajattack: It's debatable whether we want to enforce a hardware limit in software. But worth considering.

fjtrujy:
Sorry guys, the test that I did wasn't a valid test.
I was thinking that we support textures bigger than 512px, but the PSP actually didn't support it.

When we use a texture width bigger than 512 it duplicates the content of the beginning of the line.
I didn't detect that issue before because my tests weren't good enough to detect it.

If you see, fo the next example, using a `608 width`, this is how `PPSSPP` shows it (as it should be....):
Screenshot_2022-04-07_at_13 00 48

And this is how a real PSP shows it:
IMG_9285

sharkwouter:
Thanks for the update @fjtrujy. I'll close this for now, since it might cause issues for people if they develop using PPSSPP and then find out their code doesn't run on a PSP.

PR #104, opened 2022-10-03
fix format specifier

diamant3:
This PR is fixed for format specifier warnings in `exception.c` and `profiler.c` from `src/debug`. Please see these [warning logs](https://gist.github.com/diamant3/c6f818eb1b53d7fa702bd07191e74f6b).

I did not touch the `stacktrace.c` and `pspdebugkb.c` warnings because they are not the same issue.


sharkwouter (approved):
Thanks! Looks good

PR #105, opened 2022-10-14
fix restrict warning

diamant3:
This is the issue:

snprintf

Solution: 

I defined a temporary `out` buffer with the size of `PSP_DEBUG_KB_MAXLEN` and filled it with zeroes for `snprintf` to prevent truncation of the input string(`str` parameter).

### DIFF
--- a/src/debug/pspdebugkb.c
+++ b/src/debug/pspdebugkb.c
@@ -251,8 +251,8 @@ void pspDebugKbInit(char* str) {
 	  break;
 	case 1: // Space
 	  if (strlen(str) < PSP_DEBUG_KB_MAXLEN) {
-	    char out[PSP_DEBUG_KB_MAXLEN] = {0};
-	    snprintf(out, strlen(str), "%s ", str);
+	    char out[PSP_DEBUG_KB_MAXLEN + 2];
+	    snprintf(out, sizeof(out), "%s ", str);
 	    pspDebugKbDrawString(out);
 	  }
 	  break;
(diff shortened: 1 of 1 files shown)


src/debug/pspdebugkb.c
```diff
@@ -251,8 +251,9 @@ void pspDebugKbInit(char* str) {
 	  break;
 	case 1: // Space
 	  if (strlen(str) < PSP_DEBUG_KB_MAXLEN) {
-	    snprintf(str, strlen(str)+2, "%s ", str);
-	    pspDebugKbDrawString(str);
+      char out[PSP_DEBUG_KB_MAXLEN] = {0};
```
rofl0r, line 254: indentation is messed up, and you don't need to zero-initialize the buffer either. snprintf always zero-terminates the result, even if it doesn't fit into the buffer.

as second parameter for snprintf, use `sizeof(out)`.

diamant3: Thanks for the review! I tried what you have said, but it results in a format-truncation warning when I use `sizeof` to `out` buffer and remove the zero-initialization. Even if I add the `PSP_DEBUG_KB_MAXLEN` size, the result will be the same.

Sorry for the indentation, I will fix that. 😅

review1

rofl0r: then make the buffer size PSP_DEBUG_KB_MAXLEN + 2. note that that truncation is no real issue, whereas your code using strlen(str) will lead to buffer overflow and stack corruption if the str's length is >= sizeof(out). one of the many cases where someone trying to mute warnings actually introduces worse bugs.

diamant3: Thank you for your insight! I'll just close this PR to be safe.

rofl0r: well, the original warning is actually legit.

diamant3:
I try to open it again, hoping that it won't cause a buffer overflow or stack corruption because it uses a conditional statement `if (strlen(str) < PSP_DEBUG_KB_MAXLEN)` before i set the buffer, so it will check the size before doing the `snprintf`, what do you think?

rofl0r:
```c
char out[PSP_DEBUG_KB_MAXLEN+2];
snprintf(out, sizeof(out), "whatever", str);
pspdebugfoo(out);
```
you dont need to add any checks, snprintf does it already - you just need to pass the exact buffer size, not strlen.

diamant3:
Thank you! i will do that right away.

sharkwouter (approved):
Looks good to me

PR #113, opened 2023-01-26
Add the usage of triangles for textures

fjtrujy:
Improve the `speed` example using as well primitive triangles for biting the texture.

Cheers.

### DIFF
--- a/src/samples/gu/speed/speed.c
+++ b/src/samples/gu/speed/speed.c
@@ -40,9 +40,9 @@ struct Vertex
 	unsigned short u, v;
 	unsigned short color;
 	short x, y, z;
-};
+} __attribute__ ((packed));
 
-void simpleBlit(int sx, int sy, int sw, int sh, int dx, int dy)
+static void spriteBlit(int sx, int sy, int sw, int sh, int dx, int dy)
 {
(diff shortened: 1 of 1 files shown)


src/samples/gu/speed/speed.c
```diff
@@ -59,27 +59,112 @@ void simpleBlit(int sx, int sy, int sw, int sh, int dx, int dy)
 	sceGuDrawArray(GU_SPRITES,GU_TEXTURE_16BIT|GU_COLOR_4444|GU_VERTEX_16BIT|GU_TRANSFORM_2D,2,0,vertices);
 }
 
-void advancedBlit(int sx, int sy, int sw, int sh, int dx, int dy, int slice)
+static void triangleBlit(int sx, int sy, int sw, int sh, int dx, int dy)
+{
+	// simple blit, this just copies A->B, with all the cache-misses that apply
```
sharkwouter, line 64: It might be nice to include a comment explaining why there are some many vertices. I don't think that's common knowledge.

sharkwouter (commented):
It looks good and works well. I'd merge it, but since it was a bit confusing to me initially, I think it would be nice to mention why the amount of vertices changed in a comment for reference. Other than that, it's great. Thanks!

PR #114, opened 2023-01-31
Set cmake system to PSP instead of generic

sharkwouter:
Addresses this issue: https://github.com/pspdev/pspdev/issues/41

I'd like for it to be reviewed, because I do not know if it will cause any breakage yet. I'll do some testing with it.


SeleDreams:
One thing I found out is that when a name isn't generic and isn't recognised by cmake, it results in cmake displaying warnings
i wonder how the devkitpro toolchain works around this issue

sharkwouter:
I've been testing this for a bit and it is causing too many issues unfortunately.

PR #115, opened 2023-02-08
Update psputilsforkernel.h

JoseAaronLopezGarcia:
Updated fourth argument of decompression routines. When not NULL, the routines will store a pointer to the next unprocessed compressed byte (src + n_compressed_bytes_processed). This can be useful to implement stream-based decompression, calculating number of processed compressed bytes, etc.


sharkwouter (approved):
Looks good to me, thanks!

PR #119, opened 2023-03-05
Add SDL2 examples

bucanero:
While learning the basics of pspdev, I found some SDL2 examples on https://pspdev.github.io/ , but those samples where not available on pspsdk.

To help newcomers, I'm adding these files to the pspsdk samples (and to avoid copy/paste issues or mistakes)

sdl examples credits
- https://github.com/pspdev/pspdev.github.io/pull/9
- https://github.com/pspdev/pspdev.github.io/pull/12

### DIFF
--- a/src/base/pspdev.cmake
+++ b/src/base/pspdev.cmake
@@ -22,7 +22,7 @@ SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
 SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
-SET(PKG_CONFIG_EXECUTABLE "${PSPDEV}/bin/psp-pkg-config")
+SET(PKG_CONFIG_EXECUTABLE "${PSPDEV}/bin/psp-pkg-config" CACHE STRING "Path to pkg-config")
 
 ## Add Default PSPSDK Libraries according to build.mak and add stdc++ for C++ builds so this doesn't need to be done manually later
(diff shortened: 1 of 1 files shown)


sharkwouter:
Hey @bucanero, thanks for this PR. I do think it's good to have these examples available, but I would say that this is not the place, since they require specific libraries which are not available in this repository or during the testing. I think it could make sense to add them to the packages in psp-packages instead.

bucanero:
> Hey @bucanero, thanks for this PR. I do think it's good to have these examples available, but I would say that this is not the place, since they require specific libraries which are not available in this repository or during the testing. I think it could make sense to add them to the packages in psp-packages instead.

thanks for the feedback @sharkwouter 

I was checking the `psp-packages` repo, but I'm not sure where I should add these examples.
Could they be part of the SDL2 script? or do you have any other suggestions?

Just asking before opening a new PR to avoid issues

sharkwouter:
You'd add it to the directory in which the PSPBUILD for the package is located, then add it in sources in the PSPBUILD and to sha256sums you add a 'SKIP' for every file you've added. Then the package step should put the sample in a location which makes sense, probably something like `$pkgdir/psp/share/samples/$pkgname`

bucanero:
> You'd add it to the directory in which the PSPBUILD for the package is located, then add it in sources in the PSPBUILD and to sha256sums you add a 'SKIP' for every file you've added. Then the package step should put the sample in a location which makes sense, probably something like `$pkgdir/psp/share/samples/$pkgname`

great, thanks for the feedback. I've moved this PR to `draft`, and I'll create a new PR on `psp-packages` as you suggested. Once I have a new PR I'll link and close this draft.

PR #120, opened 2023-03-11
pspdev.cmake: Fix pkg-config support

glebm:
`PKG_CONFIG_EXECUTABLE` must be a cache variable, otherwise only the first call to `find_program(PkgConfig)` succeeds.


sharkwouter (approved):

sharkwouter:
Thanks for this! This should resolve at least one issue if what you said is correct

PR #125, opened 2023-04-14
build.mak: Improve and streamline .SFO creation

MotoLegacy:
The current state of building .SFO binaries is rather primitive -- as it stood, build.mak forced its own flags to be used, always. If you were building without PSP_LARGE_MEMORY, build.mak would use a deprecated tool for generation as well, one that cobbles together undocumented flags in an array of bytes and shoving it into a file.

The first change this commit makes is by enabling use of the "new" mkfsoex at all times, which creates a smaller but still bootable SFO without any arguments given. Developers are able to provide custom SFO flags now using this, added via `SFOFLAGS` in their Makefile.

This means that developers can now (at their own discretion) provide custom region information, parental control level, minimum firmware boot level, etc. with:

```make
SFOFLAGS = -d REGION=16394 -d PARENTAL_LEVEL=4 -s PSP_SYSTEM_VER="4.01"
```

As a side-effect, this also makes the SDK more adapted to new custom firmware releases that may add new SFO flags.

The second change made is turning `PSP_LARGE_MEMORY` into an opt-out /enabled by default flag if developers are targeting a firmware version newer than 3.90. Custom firmware versions starting from 4.01 M33 guard against expanding the user memory partition to 52mB if the unit detected is PSP-1000. Therefore having this check in place becomes redundant, potentially complicating Makefiles. A warning is printed out allowing developers to be aware of this fact and removing the flag from their Makefile.

### DIFF
--- a/src/base/build.mak
+++ b/src/base/build.mak
@@ -21,7 +21,7 @@ LD       = psp-gcc
 AR       = psp-gcc-ar
 RANLIB   = psp-gcc-ranlib
 STRIP    = psp-strip
-MKSFO    = mksfo
+MKSFO    = mksfoex
 PACK_PBP = pack-pbp
 FIXUP    = psp-fixup-imports
 ENC		 = PrxEncrypter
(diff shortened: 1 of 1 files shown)


src/base/build.mak
```diff
@@ -34,14 +34,28 @@ CFLAGS   := $(addprefix -I,$(INCDIR)) -G0 $(CFLAGS)
...
 
-ifeq ($(PSP_LARGE_MEMORY),1)
-MKSFO = mksfoex -d MEMSIZE=1
-endif
-
 ifeq ($(PSP_FW_VERSION),)
 PSP_FW_VERSION=150
 endif
 
+# CFW versions after M33 3.90 guard against expanding the
```
sharkwouter, line 41: What about official firmware?

src/base/build.mak
```diff
@@ -34,14 +34,28 @@ CFLAGS   := $(addprefix -I,$(INCDIR)) -G0 $(CFLAGS)
...
+# user memory partition on PSP-1000, making MEMSIZE obsolete.
+# It is now an opt-out policy with PSP_LARGE_MEMORY=0
+ifeq ($(shell test $(PSP_FW_VERSION) -gt 390; echo $$?),0)
+ifeq ($(PSP_LARGE_MEMORY),1)
+$(warning "PSP_LARGE_MEMORY" flag is not necessary targeting firmware versions above 3.90)
```
sharkwouter, line 46: Do we not want to set `MEMSIZE` at all if `PSP_LARGE_MEMORY` is set to true?

sharkwouter (commented):
This looks like a useful change, thank you for making it. I just do not understand this well enough yet to merge it, so I asked for some clarification.

src/base/build.mak
MotoLegacy: Thanks for finding this oversight, I've made a new commit that addresses it.

sharkwouter (approved):

sharkwouter:
Thanks!

Issue #46, opened 2020-11-04
Wrong struct format of SceNetAdhocctlParams


ANR2ME:
SceNetAdhocctlParams struct have a wrong format, it should have the `bssid` and `nickname` position swapped (at least on newer firmware, not sure about old firmware) otherwise games like _Ys vs. Sora no Kiseki_ will shows the wrong game room name (gibberish/random letters) taken from mac address (bssid) instead of the player's nickname when a PSP emulator is using the wrong struct format like this:
```
struct SceNetAdhocctlParams
{
	/** Channel number */
	int channel;
	/** Name of the connection */
	char name[8];
	/** The BSSID */
	unsigned char bssid[6];
	/** Nickname */
	char nickname[128];
};
```
PS: This was tested by comparing the struct contents taken from _Ys vs. Sora no Kiseki_ on JPCSP using prx files from official firmware 6.xx

Should have similar format to what VitaSdk have:
```
typedef struct SceNetAdhocctlParameter {
     int channel;
     struct SceNetAdhocctlGroupName groupName;
     struct SceNetAdhocctlNickname nickname;
     struct SceNetAdhocctlBSSId bssid;
 } SceNetAdhocctlParameter;
```

fjtrujy:
Hello @ANR2ME,
This is a really interesting question......
I have some "documentation around with version 6.60" and they show in the same format that we have, having `nickname` as the latest option.

However, if we take a look at `PPSSPP` they have :
https://github.com/hrydgard/ppsspp/blob/9271db09163d7ae3becd35c73decab0b3acad6d3/Core/HLE/proAdhoc.h#L284-L290
```c
// Active Virtual Network Information (Adhoc Group Host/Creator's device info, similar to AP?)
typedef struct SceNetAdhocctlParameter {
  s32_le channel;
  SceNetAdhocctlGroupName group_name; // This group name is probably similar to SSID name on AP
  SceNetAdhocctlNickname nickname; // According to the old PSPSDK this is the bssid, but according to the dumped content when using newer firmware this is the nickname (this is also the nickname on VitaSDK)
  SceNetAdhocctlBSSId bssid; // FIXME: bssid and nickname position might be swapped on older/newer firmware?
} PACK SceNetAdhocc	tlParameter;
```
As you can see there is a funny comment `// FIXME: bssid and nickname position might be swapped on older/newer firmware?`

So actually I'm not sure about what values we should put

ANR2ME:
[...]

PR #126, opened 2023-04-17
SceNetAdhocctlParams: Swap order of BSSID and Nickname

MotoLegacy:
Prompted by #46, I have taken the time looking at `pspnet_adhocctl.prx` in older firmware versions via Ghidra (thanks @sajattack) and some early Adhoc-capable titles like Ridge Racer and Hot Shot's Golf, I have concluded that the order of elements in the `SceNetAdhocctlParams` struct to have always been `nickname`, followed by `bssid`. Not changed in any later firmware revision. This PR resolves this discrepancy.


sharkwouter (approved):
Sounds sensible, thanks for figuring this out!

PR #130, opened 2023-05-11
ATRAC3: Fix mysterious "uint" type reference in pspatrac3.h

MotoLegacy:
Not sure where this originated. Possibly a leftover from old SDK versions since the uint type appears in some archaic homebrew as well. This was the only reference of `uint` in the entire SDK. Went ahead and changed this appropriately to `u32`. File is unable to be included into most workflows without this change.


sharkwouter (approved):

sharkwouter:
Thanks!

PR #137, opened 2023-06-23
Add dependencies for building packages to container

sharkwouter:
The pspsdk container is only used by psp-packages and psplinkusb. In the psp-packages actions we are installing the dependencies needing for building packages over 30 times. This is a waste of time and resources (it's also kinda mean towards the Alpine Linux team). My suggestion is to make sure those dependencies are already available in this container. This would probably make psp-packages build 10 times faster.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,3 +33,5 @@ ENV PSPDEV /usr/local/pspdev
 ENV PATH $PATH:${PSPDEV}/bin
 
 COPY --from=0 ${PSPDEV} ${PSPDEV}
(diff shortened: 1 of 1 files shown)


MotoLegacy (approved):
Definitely something I agree is necessary.

diamant3 (approved):

wally4000:
I don't see any issues with this.

fjtrujy:
I'm not a fan of adding these dependencies at the docker... but if it is what you wish I won't go against your thought.

fjtrujy (approved):

PR #145, opened 2023-07-10
Fix htmlviewer sample

bvdaakster:
In this PR I managed to get the htmlviewer sample working on my PSP 1000 (6.60). Looking back at the diff I don't recall exactly which change got it working but I have a feeling it's the added PSP_HEAP_SIZE_KB.

I moved the GU code to a separate file and added a simple index.html file that's loaded relative to cwd when the sample is started from memory stick. In doing so I took out the original URL to ps2dev.org, which I hope is OK.

Besides this invite for code review; if you do own any other PSP model and want to give this a test run then please share your results as well.


rofl0r:
> I don't recall exactly which change got it working

well, then maybe you try it out one by one until you figure it out?

bvdaakster:
I just checked out the original code. Indeed adding `PSP_HEAP_SIZE_KB` (or `PSP_HEAP_THRESHOLD_SIZE_KB` as I have in the last commit) fixes the issue.

diamant3 (changes requested):
It's just a suggestion, if it's ok to you, can you revert to one source file only (just the main.c file and Makefile.sample), then add the `PSP_HEAP_THRESHOLD_SIZE_KB(0x8000)` and set the url to an unsecured website like this one: `http://hitmen.c02.at/files/yapspd/?` for much cleaner.

I tested that using PSP 1000(6.60 PRO-C).

diamant3:
> Would you prefer a diff of just the fix and URL change? 

Yes, It's much cleaner without relying on those extra files.

diamant3:
LGTM👍, thanks!

sharkwouter (approved):

diamant3 (approved):

sharkwouter:
Thanks!

PR #149, opened 2023-09-15
gu: Add clang-friendly codepath

Beyley:
Clang refuses to assemble the `asm` macro, stating the instruction is "unavailable with your current feature set", and im not sure why, ive tried enabling a bunch of features, and nothing seems to make it happy. So i enabled the non-accelerated codepath on clang, which makes it compile, and allows most of gu (at least all thats needed for a cube sample) to be compiled using the LLVM/Clang 16 shipped with the Zig compiler

### DIFF
--- a/src/gu/sceGuTexImage.c
+++ b/src/gu/sceGuTexImage.c
@@ -14,13 +14,7 @@ static int tsizecmd_tbl[8] = { 0xb8, 0xb9, 0xba, 0xbb, 0xbc, 0xbd, 0xbe, 0xbf };
 
 int getExp(int val)
 {
-#ifndef __clang__
-	unsigned int i;
-	asm("clz %0, %1\n":"=r"(i):"r"(val&0x3FF));
-	return 31-i;
-#else
 	return 31 - __builtin_clz(val & 0x3FF);
-#endif
 }
 
 void sceGuTexImage(int mipmap, int width, int height, int tbw, const void* tbp)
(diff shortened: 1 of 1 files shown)


IridescentRose:
Can you not use `__builtin_clz()`?

Beyley:
> Can you not use `__builtin_clz()`?
> 
> Edit: Since you're using zig you could just export the zig-ified version of this function and use the `@clz()` builtin

I was able to use `__builtin_clz` yes, good catch

And my zig PSP SDK is trying to re-use as much of the C SDK as possible (since i dont have nearly as much time/resources/experience). Which is why im re-using the pspsdk C implementations of `gu` and `gum`, rather than using the Zig implementations provided by `zpsp-dev/zig-psp`, since those ports are outdated and would have been a large effort to bring up to Zig 0.12.0-dev
So id rather reuse the C implementations of this, since `__builtin_clz` and `@clz` will compile down to the same LLVM IR anyway

fjtrujy:
@davidgfnet could you take a look at this change?

sharkwouter:
@Beyley are you still planning on updating this with the feedback given? I'd like to be able to merge this.

Beyley:
> @Beyley are you still planning on updating this with the feedback given? I'd like to be able to merge this.

My bad, i got busy and completely forgot! I'll switch it to use __builtin_clz

Beyley:
Done

sharkwouter (approved):

sharkwouter:
Thanks, if the ci passes I'll merge this.

PR #153, opened 2023-09-26
fix resource leaks

diamant3:


sharkwouter (approved):
Thanks for the PR! Nice work on finding this oversight.

PR #154, opened 2023-09-27
fix `va_end` missing

diamant3:


sharkwouter:
I'm going to need some clarification for this one, as I'm not familiar with these functions. What happens here?

sharkwouter (approved):

PR #156, opened 2023-09-30
Add statvfs() based on sceIoDevctl()

bucanero:


sharkwouter (changes requested):
I think the build failure is because the makefile does not put `statvfs.h` in the `sys` directory.

fjtrujy:
I need to prepare newlib first.
Let me take a look

sharkwouter (approved):

PR #159, opened 2023-11-22
libcglue fix: d_type field uninitialized in getdents

irixxxx:
fix uninitialized d_type field in getdents in libc glue


fjtrujy (approved):
Thanks for the contribution, merging it

PR #167, opened 2024-01-31
pspdev.cmake: -G0 flag removed

Crow-bar:
Sorry.. I think this is the final PR

https://github.com/pspdev/psptoolchain-allegrex/issues/24


diamant3:
it's okay but let me check first this too

diamant3 (approved):

PR #169, opened 2024-02-02
rtc: use ScePspDateTime, instead of pspTime

Crow-bar:
it was a bug


sharkwouter:
Thanks! This actually fixes a bug in SDL2.

sharkwouter (approved):

PR #170, opened 2024-02-02
mp3: new API definitions

Crow-bar:
Missing APIs


sharkwouter (changes requested):
I just tested this and it gives me the following error when launching the mp3 sample eboot from the automated build:
```
ERROR: sceMp3Init returned 0x80671302
Press X to quit.
```
The file `ms0:/MUSIC/Test.mp3` exists on my system. So I think this change breaks the MP3 example.

sharkwouter:
Well, it seems I see the same error for the sample with the current build...

sharkwouter (approved):
Turns out the issue was the mp3 file I was using. It works just like the current version of the sample.

sharkwouter:
Thanks for contributing!

PR #172, opened 2024-02-02
gu: speedup

Crow-bar:
Inline expansion
Adds several frames in heavy graphics applications


sharkwouter:
It looks fine, but what makes it faster? It looks like the same code to me.

Crow-bar:
> It looks fine, but what makes it faster? It looks like the same code to me.

Are you seriously?

Crow-bar:
stack work and unnecessary jump instructions... inlining removes this

sharkwouter:
No need to be mean about it. I was expecting the compiler to be able to deal with this, I'm no expert on how the compiler works. That's why I was a bit confused and thought I had missed something while looking at the code.

I'll give this a test and merge it.

sharkwouter (approved):

sharkwouter:
Thanks for contributing, this is a really cool change!

Crow-bar:
> No need to be mean about it. I was expecting the compiler to be able to deal with this, I'm no expert on how the compiler works. That's why I was a bit confused and thought I had missed something while looking at the code.
> 
> I'll give this a test and merge it.

Sorry, i didn't mean to put it that way.
Ok, thank you

rofl0r:
not so cool are the 50 lines with gratuitous whitespace changes.

rofl0r:
> Alignment, for readability

that's nothing but a personal style preference. if done, it should be a separate commit. like this you only get a hard time seeing the actual relevant changes of the commit.

Crow-bar:
> > Alignment, for readability
> 
> that's nothing but a personal style preference. if done, it should be a separate commit. like this you only get a hard time seeing the actual relevant changes of the commit.

I agree with you

davidgfnet:
Thanks!
We should really focus on getting good LTO support. It does help in things like this. And it's (in many cases) "free" optimization :)

PR #173, opened 2024-02-04
add mp3 file sample

diamant3:


sharkwouter (changes requested):
Could you also change the location the mp3 file is loaded from? It can just be changed to "Test.mp3" and that should work.

diamant3:
donee

sharkwouter (approved):
I tested this and it works well. Thanks!

PR #175, opened 2024-02-11
Fix documentation on vertex layout for sceGuDrawArray

Sokus:
Using following Vertex layout worked:
```c
int16_t bone_weights[1-3]; // different weight count in different meshes
uint16_t texcoords[2];
uint16_t normal[3];
uint16_t position[3];
```
Even though in cases where bone weight count is 1 or 3 vertices aren't 32-bit aligned.

Two examples of layouts that would not work (and how they were fixed):
```c
int8_t weights[8];
int16_t texcoords[2];
int8_t normals[3];
// <--- adding 8-bits of padding here fixes the problem
int16_t position[3];
// <--- adding 8-bits of padding here turns (only) position into garbage
```
```c
int8_t bone_weights[1-8]; // different weight count in different meshes
// <--- 8-bit padding required here if sizeof(bone_weights) is an odd number
uint16_t texcoords[2];
uint16_t normal[3];
uint16_t position[3];
```

### DIFF
--- a/src/gu/pspgu.h
+++ b/src/gu/pspgu.h
@@ -5,7 +5,7 @@
  *
  * Copyright (c) 2005 Jesper Svennevid
  */
-      
+
 #ifndef __pspgu_h__
 #define __pspgu_h__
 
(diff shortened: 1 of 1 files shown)


sharkwouter (approved):

sharkwouter:
Thanks for contributing!

PR #188, opened 2024-03-11
Improve exec support

fjtrujy:
This PR solves issues with fork, wait and pipe.
Adding dummy implementation of `_execve`

This PR also requires to be merged: https://github.com/pspdev/newlib/pull/10

### DIFF
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -661,6 +661,14 @@ pid_t _wait(int *unused)
 }
 #endif
 
+#ifdef F__execve
+int _execve(const char *name, char *const argv[], char *const env[])
+{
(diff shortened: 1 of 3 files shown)


src/libcglue/glue.c
```diff
@@ -661,6 +661,14 @@ pid_t _wait(int *unused)
 }
 #endif
 
+#ifdef F__execve
+int _execve(const char *pathname, char *const _Nullable argv[], char *const _Nullable envp[])
```
rofl0r, line 665: not 100% sure, but i think using _Nullable here would preclude compilation with anything but the latest GCC versions.

fjtrujy: Fixed!

PR #189, opened 2024-03-12
build.mak: default to firmware 6.0

### LINKED ISSUE #89: RFC: failsafe dealing with _PSP_FW_VERSION
i've been trying to port my gnuboy fork to psp, and while it works fine in PPSSPP, it crashes on real psp with fw 6.60, which might or might not be due to wrong struct layouts when *anything* using these structs has been compiled with the wrong value for _PSP_FW_VERSION. so i recompiled pspsdk, GL, SDL2 etc with -D_PSP_FW_VERSION=600, yet it still crashes (i might have missed some library which is still using the default of 150).

which made me think of how this imo annoying situation with different struct layouts can be solved and the SDK user doesn't have to deal with it at all.
musl libc solves similar issues by having wrappers for syscalls that use structs that for some reason might be different in userspace than what the kernel expects.

let's say there's a syscall `void sceFooBar(struct sceFooBarStruct*);` and a `struct sceFooBarStruct` that would currently look like:

```c
struct sceFooBarStruct {
  int x;
#if _PSP_FW_VERSION > 150
  int foo; /* fw 1.5.1 added foo member in the middle of struct */
#endif 
  int y;
};
```
this clearly raises a lot of issues, a user of sceFooBar() that doesn't have the entire sdk and all libraries compiled with the exact firmware macro then what this will run on may either pass a struct that's too short (which will result in the kernel overwriting or reading some unrelated memory) or values intended for the y slot will end up somewhere else.

this could be dealt with by designing the struct like this:
```c
struct sceFooBarCompatStruct {
  int x;
  int y;
  int foo; /* put new member at the end */
};

struct sceFooBarStruct150 {
  int x;
  int y;
};

struct sceFooBarStructGt150 {
  int x;
  int foo;
  int y;
};
```

and then making sceFooBar() a wrapper:

```c
void sceFooBar(struct sceFooBarCompatStruct *foo) {
  struct sceFooBarStruct150 f150;
  struct sceFooBarStructGt150 later;
  if (sceKernelDevkitVersion() <= 0x01050001) {
     f150.x = foo->x;
     f150.y = foo->y;
     realSceFooBar(&f150);
  } else {
     later.x = foo->x;
     later.y = foo->y;
     later.foo = foo->foo;
     realSceFooBar(&later);
  }
}
```

from what i can tell this would completely eliminate the need to set _PSP_FW_VERSION and would make the code work on all firmware versions.
i'm not sure how the actual syscalls are done but from what i've seen from psp-fixup-imports tool it looks like stubs are used that could be used for wrapping syscalls like this.

please let me know what you think of this proposal or if you have better ideas.

rofl0r:
as discussed in #89, few structs have different layout depending on the _PSP_FW_VERSION macro.

this makes a noticable (i.e. resulting in potential memory corruption) difference only in few structs that are supposed to be used only by kernel experts, but it makes it clear to non-experts that the struct layouts they program for are the latest ones.

an expert trying to target an older firmware version, can still override the macro on the commandline.

closes #89


rofl0r:
maybe this should also be added to configure.ac, currently when pspsdk built firmware version isn't set at all...

liclac (approved):
Yeah, makes sense to me. This default dates back to an era when OS versions we're a lot more fragmented, so you wanted to target the lowest common denominator. These days, we've got CFW for the last-ever OS release, and there's very little reason not to run it.

rofl0r:
@wally4000 certainly, can you check whether it's done correctly in the last commit ? alternatively it can be added further up at `SET(CMAKE_C[XX]_FLAGS_INIT...`

fjtrujy:
Yeah, also happy with this change

diamant3 (approved):

PR #192, opened 2024-04-01
Add F_SETFD cmd to _fcntl function

fjtrujy:
## Description

This change fixes the directory iterator.

The reason why it started to fail recently is that we are every time more and more POSIX. 
After the latest changes on `newlib` and `pspsdk`, how we now support more POSIX functions the libcpp libraries are trying a new way/path (probably more optimal) to do such kind of operations.
Then this went for a path with an existing issue that we hadn't faced before.

Here you have the output using PPSSPP with the dummy project [directory-iterator-psp](https://github.com/sharkwouter/directory-iterator-psp)

Screenshot 2024-04-01 at 22 52 32

Cheers.

### DIFF
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -443,6 +443,19 @@ int _fcntl(int fd, int cmd, ...)
 			return 0;
 			break;
 		}
+		case F_SETFD:
+		{
+			int newfl;
+			va_list args;
+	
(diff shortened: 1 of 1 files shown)


diamant3 (approved):

sharkwouter:
I just tested this and it solves the problem in OceanPop. Thanks!

PR #194, opened 2024-04-08
fix include_next breakage

diamant3:
Hello, I have a fix on this issue regarding the breakage of `#include_next` because i can't compile my c++ code using Makefile, the root cause is the PR #191 I merged. To not interfere, i just created a new variable `USE_SYSTEM_LIBC`, they can set it to `1` if they want to use the `-isystem` and the default is the one without the `-isystem`, just only the Make because CMake is smarter when it comes to generation of Makefile

error:

```
psp-g++ -isystem /usr/local/pspdev/psp/include -I. -I/usr/local/pspdev/psp/sdk/include -O2 -Wall -isystem /usr/local/pspdev/psp/include -I. -I/usr/local/pspdev/psp/sdk/include -O2 -Wall -fno-exceptions -fno-rtti -D_PSP_FW_VERSION=600   -c -o main.o main.cpp
In file included from /usr/local/pspdev/psp/include/c++/13.2.0/ext/string_conversions.h:43,
                 from /usr/local/pspdev/psp/include/c++/13.2.0/bits/basic_string.h:4097,
                 from /usr/local/pspdev/psp/include/c++/13.2.0/string:54,
                 from main.cpp:1:
/usr/local/pspdev/psp/include/c++/13.2.0/cstdlib:79:15: fatal error: stdlib.h: No such file or directory
   79 | #include_next <stdlib.h>
      |               ^~~~~~~~~~
compilation terminated.
make: *** [<builtin>: main.o] Error 1
```

### DIFF
--- a/src/base/build.mak
+++ b/src/base/build.mak
@@ -27,10 +27,18 @@ FIXUP    = psp-fixup-imports
 ENC		 = PrxEncrypter
 
 # Add PSPSDK includes and libraries.
+ifeq ($(USE_SYSTEM_LIBC), 1)
 INCDIR   := $(INCDIR) . $(PSPSDK)/include
 LIBDIR   := $(LIBDIR) . $(PSPDEV)/psp/lib $(PSPSDK)/lib
 
 CFLAGS   := -isystem $(PSPDEV)/psp/include $(addprefix -I,$(INCDIR)) $(CFLAGS)
+else
+INCDIR   := $(INCDIR) . $(PSPDEV)/psp/include $(PSPSDK)/include
+LIBDIR   := $(LIBDIR) . $(PSPDEV)/psp/lib $(PSPSDK)/lib
+
+CFLAGS   := $(addprefix -I,$(INCDIR)) $(CFLAGS)
+endif # USE_SYSTEM_LIBC
+
 CXXFLAGS := $(CFLAGS) $(CXXFLAGS)
 ASFLAGS  := $(CFLAGS) $(ASFLAGS)
 


diamant3:
my bad, I forgot to update my pspsdk to master branch, let me fix the error output

rofl0r:
it's still odd that it's repeated twice. does the Makefile you invoke already set CXXFLAGS to the same values ?

rofl0r:
can you try to comment out the CXXFLAGS = line ?
it's already set to the same values by the build.mak include

rofl0r:
ok, thanks. i'll build the C++ toolchain rn and try your sample to figure out what goes wrong. i hope we can find a solution that fits everyone (without having to use an obscure flag when using build.mak)

rofl0r:
hmm, with my self-built psp-gcc 13.2.0 toolchain which uses all the bits from here (and should therefore be identical), this does not happen - the include stuiff works just as it should. what happens instead is that psp-gcc instead of psp-g++ is invoked as the linker after the successful compile, and that one obviously doesn't add -lstdc++ . i'm now investigating what causes the different behaviour - it's likely some config flag passed during gcc build, so if you could please hold on for a little more, thanks :)

rofl0r:
yeah, the build.mak as is has never been used for C++ code apparently.
it hardcodes to use LINK.c (which is an implicit rule for linking which explicitly runs the C compiler) to create the elf target. so you couldnt even override this in your Makefile by setting LD to CXX.

so this issue is certainly no regression, wouldn't it be for my isystem change, it would break at the next step.

anyway, i'm gonna go back investigating to see whether we can find a proper fix.

fjtrujy:
> yeah, the build.mak as is has never been used for C++ code apparently. it hardcodes to use LINK.c (which is an implicit rule for linking which explicitly runs the C compiler) to create the elf target. so you couldnt even override this in your Makefile by setting LD to CXX.
> 
> so this issue is certainly no regression, wouldn't it be for my isystem change, it would break at the next step.
> 
> anyway, i'm gonna go back investigating to see whether we can find a proper fix.

I have used `build.mak` with CPP for years in the RetroArch implementation.
RetroArch itself is done in C, however, there are plenty of cores that are done in CPP, and the linking always worked fine.
https://github.com/libretro/RetroArch/blob/master/Makefile.psp1#L82

fjtrujy:
> yeah, the build.mak as is has never been used for C++ code apparently. it hardcodes to use LINK.c (which is an implicit rule for linking which explicitly runs the C compiler) to create the elf target. so you couldnt even override this in your Makefile by setting LD to CXX.
> 
> so this issue is certainly no regression, wouldn't it be for my isystem change, it would break at the next step.
> 
> anyway, i'm gonna go back investigating to see whether we can find a proper fix.

Why is `build.mak` using `$(LINK.c)` instead `LD`?.
In PS2 we have 2 different `.mak` files, one for C and another one for CPP, is up to the developer to chose one or the other...

rofl0r:
> Why is build.mak using $(LINK.c) instead LD?.

it's a shorthand for `$(CC) $(LDFLAGS)` etc so it's shorter to type and read - i have actually never seen it before. but using it for brevity is the only rational explanation i can find. btw i'm close to finding out why the include order is correct in my build, but not in the official one, and i think it's because i use --target=mipsallegrexel-psp-elf instead of --target=psp - which causes gcc to have 2 different include dirs, namely PSPDEV/mipsallegrexel-psp-elf/include (used for gcc headers) and PSPDEV/psp/include - whereas using psp as target triple causes it to use the psp one for both.

rofl0r:
@fjtrujy indeed, using the mipsallegrexel... triplet the include order is indeed fixed and the reported issue doesn't happen anymore.

apart from changing all TARGET variables in the 5 build scripts, the following change to 004-pthread.sh is needed:

```
make -j $PROC_NR CC="$TARGET"-gcc all            || { exit 1; }
make -j $PROC_NR DESTDIR="$PSPDEV/$TARGET" install              || { exit 1; }
```
plus adding this snippet to the end of 005-gcc2.sh

```sh
for x in "$PSPDEV"/bin/mipsallegrexel-psp-elf* ; do
       o=$(printf "%s\n" "$(basename "$x")" | sed 's/mipsallegrexel-psp-elf/psp/')
       ln -sf "$(basename "$x")" "$PSPDEV"/bin/"$o"
done
```

rofl0r:
as for fixing the linking of c++ programs in build.mak, i could imagine that applying a patsubst to OBJS, replacing .o with .cpp and then doing some sort of file lookup with filter, filter-out, and wildcard macros, followed by setting LD to CXX if the filter returns something, could fix the issue automatically.

rofl0r:
```make
OBJS=foo.o bar.o
CXX_SRCS=$(wildcard $(patsubst %.o,%.cpp,$(OBJS)))
ifneq ($CXX_SRCS,)
LD=$(CXX)
else
LD=$(CC)
endif

all:
        echo $(CXX_SRCS)
        echo $(LD)
```

```sh
~/shm/foo $ touch foo.c bar.cpp 
~/shm/foo $ make
echo bar.cpp
bar.cpp
echo g++
g++
```

diamant3:
> i'm now investigating what causes the different behaviour - it's likely some config flag passed during gcc build, so if you could please hold on for a little more, thanks :)

thanks rofl0r, sure no worries we're open for better solution that fits for everyone, anyway, According to your investigation, this issue is likely on path issue, am i right?, how about solving this with https://gcc.gnu.org/install/configure.html but i think it's overkill to specify all the directory path, What do you think?

PS: As you can see i don't have enough skills with the tooling 😞

rofl0r:
> According to your investigation, this issue is likely on path issue, am i right?, how about solving this with https://gcc.gnu.org/install/configure.html but i think it's overkill to specify all the directory path, What do you think?

afaik there is no configure option for the gcc build to specify where it puts its "private" headers - it simply defaults to using the toolchain **triplet** - and that's where the issue lies - a triplet (which means **3** parts) is the generic way to specify a toolchain target. psp toolchain is the only toolchain i've ever seen that doesn't use a proper triplet but just "psp" - even though it has to use the *correct* triplet mipsallegrexel-psp-elf behind the scenes. i suspect this hackery was done to have a simple and short compiler name "psp-gcc".

what i would propose is to use the correct triplet for the build without hacks and then simply symlink every mipsallegrexel-psp-elf- binary to a shorter psp name, so nothing changes for the user. doing so would require about 10 changed lines in the toolchain build scripts and finally turn psp toolchain into a formally correct toolchain.

rofl0r:
hmm, the option `--with-gxx-include-dir=dirname` could eventually be used to fix the include_next issue. though i'd like to wait on a comment of @fjtrujy on the above before testing that.

fjtrujy:
For sure the triplet is wrong, this is something I have been discussing with @davidgfnet a lot
of times.
I’m totally fine changing it and using the correct triplet, I wouldn’t even put the symbolic links to psp-XXX, I know this is a breaking change, but is the standard way of using it.
I know that @davidgfnet plans to merge changes to mainstream, so probably is also the perfect timing to change and use the right triplet, what do you think?

rofl0r:
> probably is also the perfect timing to change and use the right triplet, what do you think?

sounds good. shall i prep a PR for the toolchain repo ?

> I wouldn’t even put the symbolic links to psp-XXX, I know this is a breaking change, but is the standard way of using it.

well, i'm afraid it's too interruptive, first for users, but even moreso for existing infrastructure, like CI and even 3rd party CI like in SDL project. we'd have to fix stuff left and right.

MotoLegacy:
> well, i'm afraid it's too interruptive, first for users, but even moreso for existing infrastructure, like CI and even 3rd party CI like in SDL project. we'd have to fix stuff left and right.

I agree. I think this situation is just the consequence of past mistake/oversight and we have an obligation to not cause large damage to existing projects when we can do so easily (or at all, period).

fjtrujy:
> > well, i'm afraid it's too interruptive, first for users, but even moreso for existing infrastructure, like CI and even 3rd party CI like in SDL project. we'd have to fix stuff left and right.
> 
> I agree. I think this situation is just the consequence of past mistake/oversight and we have an obligation to not cause large damage to existing projects when we can do so easily (or at all, period).

We can make the triplet change and just create  symbolic links for keeping retro-compatibility.
To use the proper triplet is mandatory if we want to merge the changes to mainstream

MotoLegacy:
> We can make the triplet change and just create  symbolic links for keeping retro-compatibility.
> 
> To use the proper triplet is mandatory if we want to merge the changes to mainstream 

This works for me, thanks! 😄

fjtrujy:
Said that, I encourage all of us anyway to stop using `gcc-` prefix everywhere and perform the needed changes too

diamant3:
I will close this now because my issue is already fixed from the revert(just add `LIBS = -lstdc++` on the Makefile to compile)

PR #197, opened 2024-04-16
Fix broken state

sharkwouter:
The isystem change seems to be causing issues and we still don't have a fix after 2 weeks. So my suggestion is to revert it for now.


sharkwouter:
@rofl0r this not not to discourage you. A new PR which adds your change back with a fix for Makefiles would still be welcome.

bucanero (approved):

rofl0r:
wow, this went down quick.

@sharkwouter: any other issue than the one that only affected gnu make in a configuration that was previously not working (build.mak + C++) ?
in other words no new issues at all ?

i held off on filing the required toolchain update so far since the other thread is still waiting for a comment from @davidgfnet - but maybe i should just go ahead anyway.

sharkwouter:
I think you can propose a solution either with a PR or a discussion here: https://github.com/pspdev/pspdev/discussions

We do want a solution which does not break existing code, which might be a bit hard to do.

PR #198, opened 2024-04-20
Fix lock api

fjtrujy:
To have thread-safe operations for all the `newlib` operations we have been required to do several PR:

- [`newlib`](https://github.com/pspdev/newlib/pull/11)
- [`psptoolchain-ee`](https://github.com/pspdev/psptoolchain-allegrex/pull/32)
- [`pspsdk`](https://github.com/pspdev/pspsdk/pull/198)

This PR skips the usage of the dummy lock API, letting us implement the specific one in the `pspsdk/libcglue`

Cheers

### DIFF
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -1099,7 +1099,7 @@ int renameat(int olddirfd, const char *oldpath, int newdirfd, const char *newpat
 {
 	// TODO: Do better implementation following https://linux.die.net/man/2/renameat
 	// for now use the same as rename
-	return _rename(oldpath, newpath);
+	return rename(oldpath, newpath);
 }
 #endif /* F_renameat  */
 
--- a/src/user/pspthreadman.h
+++ b/src/user/pspthreadman.h
@@ -589,6 +589,17 @@ int sceKernelPollSema(SceUID semaid, int signal);
  */
 int sceKernelReferSemaStatus(SceUID semaid, SceKernelSemaInfo *info);
 
+/** Attribute for lightweight mutex. */
+enum PspLwMutexAttributes
+{
+	/** The wait thread is queued using FIFO. */
+	PSP_LW_MUTEX_ATTR_THFIFO = 0x0000U,
+	/** The wait thread is queued by thread priority . */
+	PSP_LW_MUTEX_ATTR_THPRI = 0x0100U,
+	/** A recursive lock is allowed by the thread that acquired the lightweight mutex */
+	PSP_LW_MUTEX_ATTR_RECURSIVE = 0x0200U
+};
+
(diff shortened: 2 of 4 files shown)


src/libcglue/glue.c
```diff
@@ -1099,7 +1099,7 @@ int renameat(int olddirfd, const char *oldpath, int newdirfd, const char *newpat
 {
 	// TODO: Do better implementation following https://linux.die.net/man/2/renameat
 	// for now use the same as rename
-	return _rename(oldpath, newpath);
+	return rename(oldpath, newpath);
```
fjtrujy, line 1102: fix dummy warning

src/user/pspthreadman.h
```diff
@@ -589,6 +589,17 @@ int sceKernelPollSema(SceUID semaid, int signal);
  */
 int sceKernelReferSemaStatus(SceUID semaid, SceKernelSemaInfo *info);
 
+/** Attribute for lightweight mutex. */
```
fjtrujy, line 592: Adding extra documentation for lightweight mutex

diamant3 (approved):

diamant3:
Thank you for the PRs @fjtrujy

rofl0r:
it's kinda odd that the PSP kernel even ships these functions since it's inherently non-parallel with its use of cooperative multitasking, except when using the Media Engine CPU - but from what i gathered from looking at the code of uofw, for those locks it uses custom code. maybe @fjtrujy can shed some light ?

fjtrujy:
> it's kinda odd that the PSP kernel even ships these functions since it's inherently non-parallel with its use of cooperative multitasking, except when using the Media Engine CPU - but from what i gathered from looking at the code of uofw, for those locks it uses custom code. maybe @fjtrujy can shed some light ?

Not sure what specific stuff do you want to know more about?
Do you mean about recursive mutex?

fjtrujy:
> no, i mean why does the PSP OS provide mutexes, locks and semaphores when there's no real parallel execution ?

If you have a fully single thread, obviously you can’t have concurrency issues, but as soon as you start “emulating threads” (have in mind that PSP hardware is still single thread), you must have them otherwise you can suffer concurrency issues.
The swap between threads in PSP happens as soon as the current thread enters in sleep mode (tons of ways to do this, easiest on to run nanosleep function), or because interruptions with higher priorities are executed, jumping to a different part of your code… this is way sometimes you see how interruptions are disabled, in this way you can assure your execution is not going to be stop and jump to a different place… (that was the very first implementation of thread safe allocation on newlib in PSP years ago…)
Not sure if I have replied your question, but this is a really generic question in terms of “concurrency  issues“

PR #201, opened 2024-05-17
Adding Werror to compilation and solve warnings

fjtrujy:
## Description
This PR enable `-Werror` by default when compiling.

Additionally, this PR is needed to be able to support GCC 14

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -72,7 +72,7 @@ AC_SUBST(PSPSDK_INCLUDEDIR)
 AC_SUBST(PSPSDK_LIBDIR)
 
 # CFLAGS and CXXFLAGS used to build pspsdk libraries.
-PSPSDK_CFLAGS="$CFLAGS -mno-gpopt -Wall -D_PSP_FW_VERSION=600"
+PSPSDK_CFLAGS="$CFLAGS -mno-gpopt -Wall -Werror -D_PSP_FW_VERSION=600"
 PSPSDK_CXXFLAGS="$PSPSDK_CFLAGS -fno-exceptions -fno-rtti"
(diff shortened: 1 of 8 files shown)


src/audio/pspaudiolib.c
```diff
@@ -30,13 +30,6 @@ void pspAudioSetVolume(int channel, int left, int right)
   AudioStatus[channel].volumeleft  = left;
 }
 
-void pspAudioChannelThreadCallback(int channel, void *buf, unsigned int reqn)
```
fjtrujy, line 33: useless function...

configure.ac
```diff
@@ -72,7 +72,7 @@ AC_SUBST(PSPSDK_INCLUDEDIR)
 AC_SUBST(PSPSDK_LIBDIR)
 
 # CFLAGS and CXXFLAGS used to build pspsdk libraries.
-PSPSDK_CFLAGS="$CFLAGS -mno-gpopt -Wall -D_PSP_FW_VERSION=600"
+PSPSDK_CFLAGS="$CFLAGS -mno-gpopt -Wall -Werror -D_PSP_FW_VERSION=600 -Werror"
```
rofl0r, line 75: even twice ?

rofl0r: i don't think -Werror should be default. for CI it can be passed to configure. if it's always on users that have a slightly different setup or use an older compiler release can no longer compile the sdk for even the most harmless or even bogus warning. they'll have to do sed tricks on configure if they want to automate the build or patch configure manually.

fjtrujy: We have agreed on to stop thinking on legacy toolchains.

wally4000 (approved):

PR #203, opened 2024-05-18
Fix samples for gcc14

sharkwouter:
There were some samples which started to fail to build after upgrading to gcc 14. This fixes all of them.

### DIFF
--- a/src/samples/mp3/main.c
+++ b/src/samples/mp3/main.c
@@ -60,8 +60,8 @@ int SetupCallbacks(void)
 
 
 // Input and Output buffers
-char	mp3Buf[16*1024]  __attribute__((aligned(64)));
-short	pcmBuf[16*(1152/2)]  __attribute__((aligned(64)));
+unsigned char	mp3Buf[16*1024]  __attribute__((aligned(64)));
+unsigned char	pcmBuf[16*(1152/2)]  __attribute__((aligned(64)));
 
 
 // Macro to allow formatted input without having to use stdargs.h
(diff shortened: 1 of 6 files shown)


diamant3 (approved):

src/samples/mp3/main.c
```diff
@@ -60,8 +60,8 @@ int SetupCallbacks(void)
 
 
 // Input and Output buffers
-char	mp3Buf[16*1024]  __attribute__((aligned(64)));
-short	pcmBuf[16*(1152/2)]  __attribute__((aligned(64)));
+unsigned char	mp3Buf[16*1024]  __attribute__((aligned(64)));
+unsigned char	pcmBuf[16*(1152/2)]  __attribute__((aligned(64)));
```
rofl0r, line 64: this change makes the buffer half the size it originally was. intentional ?

sharkwouter: Not quite, internally it would use it as an unsigned short, but with gcc 14 this caused an error because the type checking is stricter now.

joel16: Wouldn't that mess with the sample though? I mean sure it may fix the build issues but have we tested the samples to see if these fixes break anything?

sharkwouter: I did test the sample and it does work. I tested all the ones I fixed on real hardware.

joel16: Good to hear, thanks!

PR #204, opened 2024-05-18
samples/net/simple: Run in user mode instead of kernel mode

xyzshantaram:
Kernel mode samples did not run correctly. This PR switches the sample over to run in userland, and changes some heap settings as well.
I'd be willing to fix this for the other samples as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ install-sh
 missing
 src/base/linkfile.prx
 stamp-h1
+
+.clang-format
+compile_flags.txt
--- a/src/samples/net/simple/main.c
+++ b/src/samples/net/simple/main.c
@@ -9,278 +9,242 @@
  * Some small parts (c) 2005 PSPPet
  *
  */
+#include <arpa/inet.h>
+#include <errno.h>
 #include <pspkernel.h>
-#include <pspdebug.h>
+#include <pspnet_apctl.h>
 #include <pspsdk.h>
-#include <stdlib.h>
+#include <psputility.h>
 #include <string.h>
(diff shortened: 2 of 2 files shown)


diamant3 (approved):
Thanks for the PR!

Wouter mostly fixed the samples, super thanks to him. If you find another issue in the samples, feel free to make a PR and that would be much appreciated.

joel16:
> Thanks for the PR!
> 
> Wouter mostly fixed the samples, super thanks to him. If you find another issue in the samples, feel free to make a PR and that would be much appreciated.

I believe those fixes are just build fixes for GCC 14. There may be some samples that are still borked or haven't been updated since some ancient cfw.

diamant3:
> I believe those fixes are just build fixes for GCC 14. There may be some samples that are still borked or haven't been updated since some ancient cfw.

yes you're right joel16 those fixes i said are from this PR #203. Feel free to make a PR for those who want to fix them, it's really appreciated ❤️

rofl0r:
it'd be great if you didn't mix whitespace changes with functional changes. makes it hard to see what was actually needed to fix the issue.

xyzshantaram:
Sorry, yeah, that was my bad - I had a pretty nightmarish experience trying to turn off clang-format but leave the language server on, and I'd committed my changes before realising the whitespace issue, so I settled for configuring it as close to the original as possible, but the code style was a bit all over the place.
  

  
If the maintainers would be interested I would love to set up formatting for the entire repo, perhaps even as a CI step.
  

  
My apologies once again for the trouble.
  

  

  
  
  
  
>

PR #212, opened 2024-07-03
Add MUSIC_PATH option to CreatePBP.cmake

kathrindc:
Small change to allow inclusion of an `snd0.at3` file in the PBP created by the CMake script.
Already using this change locally, thought I might as well see if it could be added here.


bucanero:
LGTM but I'd test with a sample cmake app to be sure everything still works if this `MUSIC_PATH` param is missing.

kathrindc:
Not passing a value should be fine, since it's set as part of the oneValueArgs values, which are replaced by `"NULL"` if not present, seems to work fine on my machine. (of course further testing would be good tho ^-^)

Screenshot 2024-07-03 at 19 58 26

diamant3:
Thanks for the PR!, Both PPSSPP and real PSP works.

diamant3 (approved):

PR #215, opened 2024-07-12
Add vram functions

sharkwouter:
This moves the vram related functions that were used in almost all gu samples to libgu itself. This way everyone can easily use them without using an external library.


diamant3 (approved):
approved to me, all samples are fine :D

src/gu/pspgu.h
```diff
@@ -1501,6 +1501,42 @@ void guSwapBuffersBehaviour(int behaviour);
...
+  * Allocate a draw buffer in vram
+  *
+  * Available pixel formats are:
+  *   - GU_PSM_5650 - Hicolor, 16-bit
+  *   - GU_PSM_5551 - Hicolor, 16-bit
+  *   - GU_PSM_4444 - Hicolor, 16-bit
+  *   - GU_PSM_8888 - Truecolor, 32-bit
+  *
+  * @param width - Width of the buffer, usually 512 (must be a power of 2)
+  * @param height - Height of the buffer, normally the height of the screen 272 (must be a power of 2)
```
rofl0r, line 1514: this comment contradicts itself - 272 isn't a power of 2.

sharkwouter: Fair, will fix it.

rofl0r: IIRC only the width must be power of 2, but not the height. it would be really wasteful otherwise - since the next power of 2 would be 512.

sharkwouter: That makes sense. I fixed it in this PR: https://github.com/pspdev/pspsdk/pull/217

PR #217, opened 2024-07-12
Remove incorrect part of a comment in pspgu.h

sharkwouter:


rofl0r:
there's another comment further down that repeats the "power of 2" mistake for the width param.

sharkwouter:
@rofl0r that is not a mistake, for textures it does matter. If it's not a power of 2, it will not render correctly.

diamant3 (approved):

PR #219, opened 2024-07-14
Add gu in front of vram functions

sharkwouter:
This keeps the naming convention intact.


rofl0r:
i was under the impression that gu* functions refer to kernel provided gl-like functions, as such this change might be confusing. but then i might be wrong about this.

sharkwouter:
@rofl0r I think the sceGu functions are supposed to implement features offered by the official libgu library by Sony. Some of them are a custom implementation, though, so I'm not sure.

diamant3 (approved):
Approved to me because it doesn't conflict with others, but I'm not sure too about the naming or what should be the custom inside of PSPSDK.

PR #222, opened 2024-07-18
Improve `exit` flow

fjtrujy:
## Description
This PR makes some improvements around the `exit` flow:
- Remove custom `abort` implementation as it is part of `newlib`
- Move `_exit` implementation from `libcglue` to `crt0` and `crt0_prx` even if it now is duplicated is cleaner to be there.
- The usage of `sce_newlib_nocreate_thread_in_start` now it is just on crt0 files.
- Remove the usage of `atexit` letting now the user to use that function without any additional issue.
- Other minor cleanup

Keep in mind that using directly `sceKernelExitGame` in your software is not recommended as it won't go through all the deinit process (`destructors` and `__libcglue_deinit`), so just use it unless you are not making usage of `newlib` at all, use instead always `exit` which is standard and it will call `sceKernelExitGame` at the end of the flow.

Cheers

### DIFF
--- a/src/libcglue/Makefile.am
+++ b/src/libcglue/Makefile.am
@@ -37,8 +37,6 @@ INIT_OBJS = \
 	__libcglue_init.o \
 	__gprof_cleanup.o \
 	__libcglue_deinit.o \
-	_exit.o \
-	abort.o \
 	exit.o
 
 LOCK_OBJS = \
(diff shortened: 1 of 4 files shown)


sharkwouter (approved):
This does not seems to cause any issues.

PR #209, opened 2024-06-13
add `realpath()` function

diamant3:
This PR is suggested from pspdev/psp-packages#165 and just a simple implementation to use the `realpath()` for PSP

PR #224, opened 2024-08-05
add gethostname implementation

tpimh:
PSP doesn't really has a hostname, however it has a nickname that can serve the same function. Been playing around with PSP networking recently, and the lack of this function was a bit frustrating. This is a very simple replacement function that I came up with.

### DIFF
--- a/src/libcglue/Makefile.am
+++ b/src/libcglue/Makefile.am
@@ -30,7 +30,7 @@ GLUE_OBJS = __dummy_passwd.o __psp_heap_blockid.o __psp_free_heap.o _fork.o _wai
 	_isatty.o symlink.o truncate.o chmod.o fchmod.o pathconf.o readlink.o utime.o fchown.o _getentropy.o getpwuid.o \
 	fsync.o getpwnam.o getuid.o geteuid.o basename.o statvfs.o \
 	openat.o renameat.o fchmodat.o fstatat.o mkdirat.o faccessat.o fchownat.o linkat.o readlinkat.o unlinkat.o \
-	realpath.o
+	realpath.o gethostname.o
 
 INIT_OBJS = \
 	__libpthreadglue_init.o \
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -40,6 +40,7 @@
 #include <psputils.h>
 #include <pspsdk.h>
 #include <psprtc.h>
+#include <psputility.h>
 
 #include "fdman.h"
 
@@ -1214,3 +1215,17 @@ char *realpath(const char *path, char *resolved_path)
 	return resolved_path;
 }
 #endif /* F_realpath */
+
+#ifdef F_gethostname
+int gethostname (char *__name, size_t __len) {
+	char nickname[_SC_HOST_NAME_MAX];
+	memset(nickname, 0, _SC_HOST_NAME_MAX);
+
+	if (sceUtilityGetSystemParamString(PSP_SYSTEMPARAM_ID_STRING_NICKNAME, nickname, _SC_HOST_NAME_MAX) != PSP_SYSTEMPARAM_RETVAL_FAIL) {
+		strlcpy(__name, nickname, __len);
+		return 0;
+	}
+
+	return __set_errno(EINVAL);
+}
+#endif /* F_gethostname */


tpimh:
Oops, psputils.h and psputility.h are not the same thing...

sharkwouter:
I did a little test with this and it does work. Here is my testing code.

main.c:
```
#include <pspkernel.h>
#include <pspdebug.h>
#include <pspdisplay.h>
#include <sys/unistd.h>

PSP_MODULE_INFO("PSP HOSTNAME", 0, 1, 0);
PSP_MAIN_THREAD_ATTR(THREAD_ATTR_USER);

int main(int argc, char *argv[])
{
    pspDebugScreenInit();

    char hostname[128];
    gethostname(hostname, 128);

    pspDebugScreenPrintf("%s", hostname);

    while (1)
    {
        sceDisplayWaitVblankStart();
    }
}
```

CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.0)

project(hostname)

add_executable(${PROJECT_NAME}
  main.c
)

target_link_libraries(${PROJECT_NAME}
  pspdebug
  pspdisplay
  pspge
)

create_pbp_file(
    TARGET ${PROJECT_NAME}
    ICON_PATH NULL
    BACKGROUND_PATH NULL
    PREVIEW_PATH NULL
    TITLE ${PROJECT_NAME}
)
```

I'm not sure if we should be using the nickname, though. I think it would probably make more sense to just set it to `PSP` or maybe `PSP-nickname` or something like that.

tpimh:
Hmmm, I need to test it. I got the same code working in my project, so I must've made a mistake when copying it to libcglue.c.

sharkwouter:
It actually works, I just thought using the nickname is a bit odd and I added a code I used to test to make it a bit easier.

tpimh:
I wonder what is the shortest and the longest nickname that can be set? And how does it compare to `_SC_HOST_NAME_MAX`?

sharkwouter:
I think the limit is 128 characters, but I don't think anyone would actually use a nickname that long.

diamant3:
Thanks for the PR @tpimh and testing @sharkwouter! 

I'll also add my review a bit to this one because we're on the same page as my PR #209 and I'll base my judgment on that.

src/libcglue/glue.c
```diff
@@ -1213,4 +1214,18 @@ char *realpath(const char *path, char *resolved_path)
 
 	return resolved_path;
 }
-#endif /* F_realpath */
\ No newline at end of file
+#endif /* F_realpath */
+
+#ifdef F_hostname
+int gethostname (char *__name, size_t __len) {
```
diamant3, line 1220: This should be the same like `F_realpath` and `realpath` for example, but the purpose is that I'm not really sure if it's to prevent naming collisions, just naming conventions, or to detect that function in the newlib.

src/libcglue/glue.c
```diff
@@ -1213,4 +1214,18 @@ char *realpath(const char *path, char *resolved_path)
...
+	char nickname[_SC_HOST_NAME_MAX];
+	memset(nickname, 0, _SC_HOST_NAME_MAX);
+
+	if (sceUtilityGetSystemParamString(PSP_SYSTEMPARAM_ID_STRING_NICKNAME, nickname, _SC_HOST_NAME_MAX) != PSP_SYSTEMPARAM_RETVAL_FAIL) {
+		strlcpy(__name, nickname, __len);
+		return 0;
+	}
+
+	return 1;
```
diamant3, line 1229: In my PR #209, I just based my implementation to the [description part here](https://linux.die.net/man/3/realpath) and If it's working and doesn't conflict with others, then I don't really have anything to say about that.

src/libcglue/glue.c
```diff
@@ -1213,4 +1214,18 @@ char *realpath(const char *path, char *resolved_path)
...
+}
+#endif /* F_hostname */
```
diamant3, line 1231: this should be the same with the name of `#ifdef` part

src/libcglue/Makefile.am
```diff
@@ -30,7 +30,7 @@ GLUE_OBJS = __dummy_passwd.o __psp_heap_blockid.o __psp_free_heap.o _fork.o _wai
 	_isatty.o symlink.o truncate.o chmod.o fchmod.o pathconf.o readlink.o utime.o fchown.o _getentropy.o getpwuid.o \
 	fsync.o getpwnam.o getuid.o geteuid.o basename.o statvfs.o \
 	openat.o renameat.o fchmodat.o fstatat.o mkdirat.o faccessat.o fchownat.o linkat.o readlinkat.o unlinkat.o \
-	realpath.o
+	realpath.o hostname.o
```
diamant3, line 33: `gethostname.o` too

tpimh:
I was thinking about implementing `sethostname` later which can be added to the same object file (`hostname.o`), but it can also be in a separate object file.

diamant3:
> I was thinking about implementing `sethostname` later which can be added to the same object file (`hostname.o`), but it can also be in a separate object file.

Sure! Feel free to make a separate PR for that. Also, it's much better that every function is separate and corresponds to each object file. Please stick with that.

src/libcglue/glue.c
```diff
@@ -1213,4 +1214,18 @@ char *realpath(const char *path, char *resolved_path)
 
 	return resolved_path;
 }
-#endif /* F_realpath */
\ No newline at end of file
+#endif /* F_realpath */
+
+#ifdef F_hostname
```
fjtrujy, line 1219: use here `F_gethostname`

src/libcglue/Makefile.am
fjtrujy: use `gethostname.o`

src/libcglue/glue.c
fjtrujy: shouldn't we return the proper errno code?

tpimh: Yes, that's a good suggestion. I should check what the proper code is.

sharkwouter (approved):
This works great and looks good now, thanks!

src/libcglue/glue.c
```diff
@@ -1213,4 +1214,18 @@ char *realpath(const char *path, char *resolved_path)
...
 	return resolved_path;
 }
-#endif /* F_realpath */
\ No newline at end of file
+#endif /* F_realpath */
+
+#ifdef F_gethostname
+int gethostname (char *__name, size_t __len) {
+	char nickname[_SC_HOST_NAME_MAX];
+	memset(nickname, 0, _SC_HOST_NAME_MAX);
```
rofl0r, line 1222: this memset looks superfluous, in the worst case if sceUtilityGetSystemParamString requires a valid C string as input, which seems unlikely, `nickname[0] = 0;` should be sufficient.

PR #228, opened 2024-08-15
Clean up some methods

fjtrujy:
## Description
A small PR where we make inline some of the not public methods.

Additionally, some minor improvements have been made too.

Cheers!


sharkwouter:
There are still some errors in the CI. I need some extra time to do some testing with the changes in `src/gu/sceGuTexImage.c`.

src/gu/sceGuInit.c
```diff
@@ -235,6 +235,25 @@ static unsigned int __attribute__((aligned(16))) ge_init_list[] =
...
 };
 
+static void callbackFin(int id, void* arg)
+{
+	GuSettings* settings = (GuSettings*)arg;
+	if (settings->fin)
+		settings->fin(id & 0xffff);
+}
+
+static void callbackSig(int id, void* arg)
```
sharkwouter, line 245:
```suggestion
void callbackSig(int id, void* arg)
```
This cannot be static, otherwise it won't match the definition in `guInternal.h`.

src/gu/sceGuInit.c
```diff
@@ -235,6 +235,25 @@ static unsigned int __attribute__((aligned(16))) ge_init_list[] =
 	ZV(END),
 };
 
+static void callbackFin(int id, void* arg)
```
sharkwouter, line 238:
```suggestion
void callbackFin(int id, void* arg)
```
This cannot be static, otherwise it won't match the definition in `guInternal.h`.

sharkwouter (changes requested):
I took a look and found out where the build errors are coming from. When you apply my suggestions, it works just fine.

sharkwouter (approved):

PR #235, opened 2024-09-09
remove old README file

sblondon:
`README.old`file  contains obsolete information (like svn repository) and has not been modified since 4 years.


joel16:
I believe the reason it still exists is to preserve history. Hence why it was renamed "README.old"

sharkwouter:
It seems like all the information in the old readme is still in the new one, apart from links to websites that do not exist anymore and the old installation instructions that no longer apply. I would say it is fair to remove it. It will still be in the git history for the people who care about older versions.

diamant3 (approved):
Thanks, I will merge this now.

PR #237, opened 2024-10-01
Make analog stick values more readable in basic control sample

sblondon:
When the SceCtrlData.Lx value changes, the length of the value can change (from 1 character to 3 characters) so the string for Ly moves to the right by one or two characters.
The SceCtrlData.Lx and Ly are `unsigned char` so the value are between 0 (one character length) and 255 (3 characters length).
It's more readable by fixing the size for the value to 3 characters wide so the string to Ly does not move.

### DIFF
--- a/src/samples/controller/basic/main.c
+++ b/src/samples/controller/basic/main.c
@@ -74,8 +74,8 @@ int main(void)
 
 		sceCtrlReadBufferPositive(&pad, 1);
 
-		pspDebugScreenPrintf("Analog X = %d ", pad.Lx);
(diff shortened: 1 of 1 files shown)


sblondon:
I have written the same PR on [pspdev.github.io](https://github.com/pspdev/pspdev.github.io/pull/64) repository.

sharkwouter (approved):

sharkwouter:
Thanks for this change, this looks better. I've changed the title of the PR because we now use them in our changelogs for releases. I hope it's okay for you like that.

sblondon:
No problem, the title is clearer now!

PR #238, opened 2024-10-11
Display build date instead of unchanged version number

sharkwouter:
Right now the version number on https://pspdev.github.io/pspsdk is always the same and we never update it. I thought it would make more sense to have the build date there, like we do with PSPDEV releases as a whole.


sharkwouter:
The result of this change can be seen here: https://sharkwouter.github.io/pspsdk/

diamant3 (approved):
Nice PR wouter, approved to me!

PR #239, opened 2024-10-23
Fix: Use sce_stat->st_mode in io_to_posix_mode (glue.c) so it can also work for files stored in ISOs

radusalagean:
Created a PR as suggested by @sharkwouter 
Key messages that led to this change which fixed my issue:

1. Initial issue explained: https://discord.com/channels/479828644970364928/668436284800237694/1298297311235145878

2. Testing `sceIoGetstat` directly:
https://discord.com/channels/479828644970364928/668436284800237694/1298625671383351296
and
https://discord.com/channels/479828644970364928/668436284800237694/1298651722356494378

3. Found a forum post which suggests that using `st_attr` does not work when checking resources from a UMD, and `st_mode` should be used instead:
https://discord.com/channels/479828644970364928/668436284800237694/1298668881333846047

4. Issue is solved for me with this change from the PR, the SDL project works on both ISO and non-ISO context: 
https://discord.com/channels/479828644970364928/668436284800237694/1298700049257201705

### DIFF
--- a/src/libcglue/glue.c
+++ b/src/libcglue/glue.c
@@ -282,11 +282,11 @@ static time_t psp_to_posix_time(ScePspDateTime psp_time)
 static mode_t io_to_posix_mode(SceMode sceMode)
 {
         mode_t posixmode = 0;
-        if (sceMode & FIO_SO_IFREG) posixmode |= S_IFREG;
-        if (sceMode & FIO_SO_IFDIR) posixmode |= S_IFDIR;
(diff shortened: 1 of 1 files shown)


bucanero (approved):
LGTM
@fjtrujy  might have a better overview

sharkwouter (approved):
So I tested this with OceanPop and I ran the following samples:
- mp3
- gu/sprite
- savedata/utility

They all seemed to work fine.

fjtrujy:
Thanks for the PR.
I will try to run some other examples just to be sure.
Thanks!

fjtrujy (approved):

PR #242, opened 2024-10-30
Add add_prx_module macro to psp-cmake

sharkwouter:
This adds a macro for creating prx modules which looks a bit like the add_executable function in cmake.


fjtrujy (approved):
Amazing

PR #243, opened 2024-10-31
Use function with easier input for generating PRX module files

sharkwouter:
This would make using this a bit easier. Right now you have to do this:
```cmake
add_prx_module(TARGET myprx
    SRC_FILES main.c extra.c
    EXP_FILES main.exp extra.exp
)
```
This change would allow you to do this instead:
```cmake
add_prx_module(myprx
    main.c
    extra.c
    main.exp
    extra.exp
)
```
The function will figure out which of the files specified are exp files and generate c files from them without the user needing to make a distinction.


src/base/AddPrxModule.cmake
```diff
@@ -0,0 +1,55 @@
...
+  endforeach()
+
+  add_executable(${name} ${SRC_FILES})
+  target_link_options(${name} PRIVATE -Wl,-q,-T${PSPDEV}/psp/sdk/lib/linkfile.prx -nostartfiles -Wl,-zmax-page-size=128)
+
+  add_custom_command(
+    TARGET ${name}
+    POST_BUILD COMMAND
+    "$ENV{PSPDEV}/bin/psp-fixup-imports" "$<TARGET_FILE:${name}>"
+    COMMENT "Calling psp-fixup-imports"
```
fjtrujy, line 44: could you put on the comments the `$<TARGET_FILE:${name}>` too?

To see things as:
```
Calling psp-fixup-imports for XXX
```

src/base/AddPrxModule.cmake
```diff
@@ -0,0 +1,55 @@
...
+
+      # Define the output .c file path
+      set(GENERATED_C_FILE ${CMAKE_BINARY_DIR}/${EXP_FILE_NAME}.c)
+
+      # Add a custom command for each .exp file to generate a .c file
+      add_custom_command(
+          OUTPUT ${GENERATED_C_FILE}
+          COMMAND psp-build-exports -b ${EXP_FILE} > ${GENERATED_C_FILE}
+          DEPENDS ${EXP_FILE}
+          COMMENT "Generating ${EXP_FILE_NAME}.c from ${EXP_FILE_NAME}.exp"
```
fjtrujy, line 26: could you put on the comments the `$<TARGET_FILE:${name}>` too?

To see things as:
```
Generating XXXX.c from XXXX.exp for YYYY
```

fjtrujy (approved):
Good improvement!

PR #244, opened 2024-10-31
Add support for setting the output directory to create_pbp_file macro

sharkwouter:
This will allow us to make sure SDL can build with `-j8` for when building tests and it is also just nice to be able to set the output directory.


diamant3 (approved):
Nice PR!, we can now set the output of EBOOT.PBP.

PR #246, opened 2024-10-31
Add missing psptypes.h

wally4000:
u8 type is not declared in pspwlan.h


diamant3 (approved):

sharkwouter:
I think it would make more sense to change the u8 use with uint8_t

sharkwouter (approved):

joel16:
> I think it would make more sense to change the u8 use with uint8_t

This change makes it inconsistent with the other headers. Most headers use the types from psptypes.h.

PR #247, opened 2024-11-03
disable autoconf default compilation flags

diamant3:
I think it's much better if we can see the CFLAGS value and can modify it.

ref: https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html#index-AC_005fPROG_005fCC-1

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,6 +13,10 @@ AC_PSPDEV_PATH
 AM_INIT_AUTOMAKE([1.14 foreign])
 AM_SILENT_RULES([yes])
 
+# Disable autoconf default compilation flags
+: ${CXXFLAGS=""}
+: ${CFLAGS=""}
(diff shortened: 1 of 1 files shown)


rofl0r:
imo not such a good idea because the user can no longer override the -g -O2 with custom CFLAGS.
compiler options coming latter override previous ones, so a user doing CFLAGS="-O3 -g0" ./configure can no longer get the desired setting.

diamant3:
Yes, you're right. it doesn't override the explicit value. I'll draft it and test it more.

rofl0r:
then the question arises "why change something that isn't broken"

diamant3:
I will close this one because it's unnecessary to specify this flag because it's always `-g` `-O2` are included.

PR #248, opened 2024-11-14
add other sceDisplay functions

diamant3:
These are the last functions found in sceDisplay from https://spenon-dev.github.io/PSPLibDoc/, but these are all untested. Can we add them? thanks.

### DIFF
--- a/src/display/Makefile.am
+++ b/src/display/Makefile.am
@@ -10,7 +10,7 @@ CPPFLAGS = -I$(top_srcdir)/src/base -I$(top_srcdir)/src/kernel
 CFLAGS = @PSPSDK_CFLAGS@
 CCASFLAGS = $(CFLAGS)
 
-DISPLAY_OBJS = sceDisplay_0000.o sceDisplay_0001.o sceDisplay_0002.o sceDisplay_0003.o sceDisplay_0004.o sceDisplay_0005.o sceDisplay_0006.o sceDisplay_0007.o sceDisplay_0008.o sceDisplay_0009.o sceDisplay_0010.o sceDisplay_0011.o sceDisplay_0012.o sceDisplay_0013.o sceDisplay_0014.o sceDisplay_0015.o sceDisplay_0016.o sceDisplay_0017.o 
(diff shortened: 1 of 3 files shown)


sharkwouter:
I did a quick test to see if this interfered with existing functions, but I had no issue. I would like to see at least one or two test applications to make sure at least some of them work, though. I think `sceDisplayGetBrightness` and `sceDisplayIsVsync` should be quite easy to test.

sharkwouter:
I did some more testing and it seems that these are wrong unfortunately. The `sceDisplayIsVsync` function actually returns a counter which starts when the application is started which keeps increasing. `sceDisplayGetBrightness` returns 0 regardless of screen brightness.

sharkwouter:
The `sceDisplayGetResumeMode` function causes the application to crash. The `sceDisplayAdjustAccumulatedHcount` function just returns 0, dunno what that means.

sharkwouter (changes requested):
Right now these functions don't do what the titles say. Some don't even work. We should fix this first.

joel16:
Well for one, I doubt all these functions have void arguments. For example sceDisplayGetBrightness seems to have two parameters from what I remember:

https://github.com/joel16/CMFileManager-PSP/blob/next/display_driver/display_driver.c#L13 (This is the kernel mode equivalent though which I assume would be similar/same in usermode)

diamant3:
> Right now these functions don't do what the titles say. Some don't even work. We should fix this first.

Got it, and if no error codes returned, it means it's just a matter of function signature. I'll draft it and continue to work on this one. Thank you guys.

diamant3:
Done. I checked them all one function at a time using my PSP slim(2000) and everything is usable except the `sceDisplayGetResumeMode`(I wrote a warning on this function).

Thanks to [Joel16' CMFileManager display driver implementation](https://github.com/joel16/CMFileManager-PSP/blob/next/display_driver/display_driver.c), [PPSSPP sceDisplay implementation](https://github.com/hrydgard/ppsspp/blob/master/Core/HLE/sceDisplay.cpp) and [vita SDK docs](https://docs.vitasdk.org/).

joel16:
I might be completely wrong, and this is nothing but a mere speculation but sceDisplayGetResumeMode sounds like something for psp go's resume feature.

diamant3:
> I might be completely wrong, and this is nothing than a mere speculation but sceDisplayGetResumeMode sounds like something for psp go's resume feature.

That function is specific to psp go? Hopefully not.😢

Anyway, Is it much better if I remove that `sceDisplayGetResumeMode` function to be safe?

joel16:
> > I might be completely wrong, and this is nothing than a mere speculation but sceDisplayGetResumeMode sounds like something for psp go's resume feature.
> 
> That function is specific to psp go? Hopefully not.😢
> 
> Anyway, Is it much better if I remove that `sceDisplayGetResumeMode` function to be safe?

Unless we can get it tested somehow then I'd suggest we remove it.

fjtrujy:
What's the status of this PR?

sharkwouter:
> What's the status of this PR?

We haven't figured out how to make this one work, so we can't merge it.

diamant3:
Hello guys, sorry I forgot about this due to lack of PSP that could test all of these functions. If anyone can take over, let me know and I will close this, thank you.

PR #251, opened 2024-12-03
Align memory to 16 when giving out vram for textures and buffers

sharkwouter:
Right now when you use `guGetStaticVramTexture` you might see rendering issues because of an alignment mismatch. This PR aligns it to prevent this issue.

### DIFF
--- a/src/gu/vram.c
+++ b/src/gu/vram.c
@@ -9,8 +9,9 @@
 #include <pspge.h>
 #include <pspgu.h>
 
+#define ALIGNMENT 16
+
 static unsigned int staticOffset = 0;
-static const unsigned int alignment = 16;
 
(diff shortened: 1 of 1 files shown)


src/gu/vram.c
```diff
@@ -10,6 +10,7 @@
 #include <pspgu.h>
 
 static unsigned int staticOffset = 0;
+static const unsigned int alignment = 16;
```
rofl0r, line 13: probably more efficient to use `#define ALIGNMENT 16` here, as C afaik doesn't evaluate constant expressions involving variables at compile time, as C++ would.

sharkwouter: Ah, I wasn't aware. I'll change it.

diamant3 (approved):

PR #258, opened 2025-01-26
samples: Add kubrdige sample

joel16:
A short and sweet sample that demonstrates how to use a kernel function inside a usermode application. This involves using kubrdige to load the kernel module, and then using its exported kernel function inside a usermode application.

### DIFF
--- a/src/samples/kernel/userbridge/user/Makefile.sample
+++ b/src/samples/kernel/userbridge/user/Makefile.sample
@@ -1,0 +1,18 @@
+TARGET = KUBridgeSample
+OBJS = main.o kernel.o
+
+INCDIR = 
+CFLAGS = -O2 -G0 -Wall
+CXXFLAGS = $(CFLAGS) -fno-exceptions -fno-rtti
+ASFLAGS = $(CFLAGS)
+
+LIBDIR =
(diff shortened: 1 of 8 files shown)


src/samples/kernel/userbridge/user/Makefile.sample
```diff
@@ -0,0 +1,18 @@
...
+
+INCDIR = 
+CFLAGS = -O2 -G0 -Wall
+CXXFLAGS = $(CFLAGS) -fno-exceptions -fno-rtti
+ASFLAGS = $(CFLAGS)
+
+LIBDIR =
+LDFLAGS =
+
+LIBS = -lpspkubridge
```
sharkwouter, line 12: This won't work in this rep unfortunately. You'd have to add the sample to the package in the psp-packages repo. We did the sample with th sdl2 sample.

joel16: Ah okay, I'll have a look at that and do the same then.

sharkwouter: Cool, thanks for the effort. It'd be great to have this.

PR #265, opened 2025-03-28
syscon: Use known/documented function names

joel16:
Based on reverse-enginnered code by @Felix-Dev from uOFW (https://github.com/uofw/uofw/blob/power/include/syscon.h). I have also added punctuation on some of the existing documentation.

### DIFF
--- a/src/kernel/pspsyscon.h
+++ b/src/kernel/pspsyscon.h
@@ -57,24 +57,33 @@ int sceSysconCtrlLED(int SceLED, int state);
  */
 int sceSysconCtrlHRPower(int power);
 
+/**
+ * Get the headphone connection status.
+ * @return 1 if the headphone is connected, 0 if the headphone is disconnected.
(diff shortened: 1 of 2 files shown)


sharkwouter:
Looks good, but do you have some code that would allow for this to be tested?

joel16:
> Looks good. Do you have some code that would allow for this to be tested?

I don't exactly have a sample, but I do use a lot of these in PSP-Everest: https://github.com/joel16/PSP-Everest

rofl0r:
maybe could use explanations about what pommel and baryon are ?

joel16:
> maybe could use explanations about what pommel and baryon are ?

IMO I don't think that fits in SDK documentation. The wiki has that information:

Baryon: https://www.psdevwiki.com/psp/Baryon
Pommel: https://www.psdevwiki.com/psp/Pommel

PR #266, opened 2025-03-29
fix: remove two warnings during documentation generation

sblondon:
Doxygen displays two warnings (among others) because parameters filled in @params are not correct.

The warnings fixed are:

```
pspsdk/src/user/pspthreadman.h:622: warning: argument 'optionsPTr' of command @param is not found in the argument list of sceKernelCreateLwMutex(SceLwMutexWorkarea *workarea, const char *name, SceUInt32 attr, int initialCount, u32 *optionsPtr) pspsdk/src/user/pspthreadman.h:622: warning: The following parameter of sceKernelCreateLwMutex(SceLwMutexWorkarea *workarea, const char *name, SceUInt32 attr, int initialCount, u32 *optionsPtr) is not documented:
  parameter 'optionsPtr'
pspsdk/src/user/pspthreadman.h:665: warning: argument 'name' of command @param is not found in the argument list of sceKernelUnlockLwMutex(SceLwMutexWorkarea *workarea, int lockCount)
```

There is another type of warning about the source code documentation like:

```
pspsdk/src/prof/pspprof.h:28: warning: argument 'filename' of command @param is not found in the argument list of __attribute__((__no_instrument_function__, __no_profile_instrument_function__))
pspsdk/src/prof/pspprof.h:28: warning: argument 'should_dump' of command @param is not found in the argument list of __attribute__((__no_instrument_function__, __no_profile_instrument_function__))
/
```
There is the same case for prof.c:278.

The issue is the `__attribute__()` before the function definition. Moving the `__attribute__()` line above the comments does not solve the issue. So I did not fixed them.

Do you have any idea how to fix them?

PR #271, opened 2025-04-03
doc: search improvement and link to constants

sblondon:
This patch fixes the search for function like `__mcount()` or `gprof_start()`. The issue existed for function with function attribute `(__attribute__())`: Doxygen didn't parsed it properly. It fixes such warnings:

```
warning: documented symbol '__mcount' was not declared or defined.
```

After changing Doxygen configuration to fix it, new warnings appears. For example: 

```
warning: explicit link request to 'PSP_VAUDIO_FORMAT_' could not be resolved
```

It's fixed by using the full name of the constants. By the way, it also fixes the link to the constants in the documentation.

Initial patch based on:
https://stackoverflow.com/questions/75410662/doxygen-does-not-parse-attribute-correctly

The first discussion about this issue is in https://github.com/pspdev/pspsdk/pull/266. The attempt with `@fn` was not satisfying because the search didn't work properly with such patch.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -2085,7 +2085,7 @@ ENABLE_PREPROCESSING   = YES
 # The default value is: NO.
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
-MACRO_EXPANSION        = NO
+MACRO_EXPANSION        = YES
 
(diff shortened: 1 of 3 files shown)


diamant3 (approved):
Neat fix, thanks!

sharkwouter:
Cool, that is a good catch!

PR #275, opened 2025-04-10
Move generated documentation in dedicated directory

sblondon:
The documentation generated by Doxygen moves from `doc/` to `doc/build/`.
There is a problem with using `doc/` directory because when Doxygen produces the documentation, the `doc/` directory is replaced so its content (`style.css` file) disappears.


sharkwouter (approved):
Looks good and works well, thanks!

PR #276, opened 2025-04-10
Improve `sceGu`

fjtrujy:
## Description

This PR does: 
- It removes some remaining magic numbers from `sceGu` functions, and instead uses the macro defined.
- It also creates a funnel for the `sceGuFinishId` function.
- Replace in all the repo the magic numbers for `sceGuSync` and `sceGuStart`


sharkwouter (approved):
Much better, thanks for the PR!

Issue #45, opened 2020-10-05
Busted Samples


sajattack:
Tracking issue for broken samples (of which there are many)
Checking the box indicates the sample has been fixed.
Please comment on this issue rather than filing a separate one
- [ ] debug/profiler

John-K:
- [ ] debug/sio

ababaei:
PSP 1000 6.60 ME-2.3 cant get to run :

- [ ] storage/usb

fjtrujy:
hello guys,
Could you give a try to the samples after the latest toolchain upgrade? maybe now they are fixed.

Thanks

mgood7123:
psp 1000 6.60 pro

Prx loader
Prx template
Kdumper
Profiler

All don't work

Fewnity:
Psp 1000 6.61 LME-2.3

I'm using the lastest psp toolchain.

utility/htmlviewer: Prints "Error 0x80020190 in sceNetInit"
utility/netdialog: Always gives an error when connecting to a wifi
net/wlanscan_elf: Crashes
net/resolver: Crashses
net/simple: Crashses

(My psp wifi works fine, I can use the web browser)

diamant3:
I just compiled all the status of samples; maybe it can help you guys.

#### Device: PSP 6.60(PRO-C)

for those **(did not test)** status, I didn't test because there is no `EXTRA_TARGETS = EBOOT.PBP`

- Audio
  - [x] polyphonic
  - [x] wavegen
- Controller
  - [x] Basic
- Debug
  - [x] debugkb
  - [ ] exception - black screen then exit with `The game could not be started (8002013c)`
  - [ ] gdb - black screen then exit with `The game could not be started (8002013c)`
  - [ ] kprintf - black screen then exit with `The game could not be started (8002013c)`
  - [ ] profiler - Stuck black screen
  - [ ] prxdecrypt - black screen then exit with `The game could not be started (8002013c)`
  - [ ] sio - black screen then exit with `The game could not be started (8002013c)`
- [x] Exceptions
- GU
  - [x] beginobject
  - [x] blend
  - [x] blit
  - [x] celshading
  - [x] clut
  - [x] common
  - [x] copy
  - [x] cube
  - [x] envmap
  - [x] integerdrawing
  - [x] lights
  - [x] lines
  - [x] logic
  - [x] mipmapping - working but cannot exit
  - [x] morph
  - [x] morphskin
  - [x] ortho
  - [x] reflection
[...]

PR #278, opened 2025-04-16
refacto: same constant declaration become unique

sblondon:
It's clearer to declare 'decrypted' outside the conditional compilation. Only the two other declarations are different.

I found it when I compiled some examples listed in [broken samples](https://github.com/pspdev/pspsdk/issues/45#issuecomment-1630213968).


sharkwouter:
Thanks! That's a nice improvement.

PR #279, opened 2025-04-19
added 'data.psar' handling to create_pbp_file

TheHellcat:
Added the ability to add a path for a "data.psar" file, to be added to the built EBOOT, when using CMake.


bucanero (approved):

sharkwouter:
I'm thinking if this is the way we want to go about adding support for psar files. I think I like the Vita approach of including files into the archive, this looks like this for them: https://github.com/sharkwouter/oceanpop/blob/75bb9856ed9f59b4a6e3429fff956818cd4bc8b0/CMakeLists.txt#L260

What do you think @TheHellcat?

sharkwouter:
The Vita guys add files here: https://github.com/vitasdk/vita-toolchain/blob/77f17b979678e7af0acfe708dc97a45ca6fd55e2/cmake_toolchain/vita.cmake#L370

PR #281, opened 2025-05-01
clean: remove unused macros

sblondon:
The macros are not used.

In `blend.c`, the values are not used except for `RING_RADIUS`: `1.0f` is used in [`sceGuTexScale()`](https://github.com/pspdev/pspsdk/blob/46482f38fd879d8cad94bc9d67557b10501d2d70/src/samples/gu/blend/blend.c#L156) call for its two parameters. However, I doubt there is a link. If you think it's the case, I can modify the PR to keep `RING_RADIUS` and use it in `sceGuTexScale()` call.

### DIFF
--- a/src/samples/gu/blend/blend.c
+++ b/src/samples/gu/blend/blend.c
@@ -45,12 +45,6 @@ struct BlendState
 #define SCR_WIDTH (480)
 #define SCR_HEIGHT (272)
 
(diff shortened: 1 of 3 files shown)


bucanero (approved):

sharkwouter:
Thanks!

PR #285, opened 2025-05-08
Some clean ups around `sceGu`

fjtrujy:
## Description
Some improvements and clean-ups in `sceGu`:

- Putting better constants names for `PspDisplaySetBufSync`
- Separate into 2 files `sceGuFinish` and `sceGUFinishId`
- Modify `sendCommandiStall` function, it was previously adding the command and then updating the stall address, now we have a new function called `_sceGuUpdateStallAddr` which updates the stall address.
- Add return value to `sceGuInit`
- Add return value to `sceGuCallList`
- Rename `resetValues` to `_sceGuResetGlobalVariables`, which is a more appropriate name for a "kind of private function"
- Remove some remaining magic numbers.
- Improve `sceGuInit` by being sure we cover failure scenarios and make it more robust for concurrency scenarios.
- Improve `sceGuSendList`
- Some other minor changes

Cheers!

### DIFF
--- a/src/display/pspdisplay.h
+++ b/src/display/pspdisplay.h
@@ -31,12 +31,15 @@ enum PspDisplayPixelFormats {
 };
 
 enum PspDisplaySetBufSync {
-	/** Buffer change effective immediately */
-	PSP_DISPLAY_SETBUF_IMMEDIATE = 0,
-	/** Buffer change effective next frame */
-	PSP_DISPLAY_SETBUF_NEXTFRAME = 1
+	/** Buffer change effective next hsync */
+	PSP_DISPLAY_SETBUF_NEXTHSYNC = 0,
+	/** Buffer change effective next vsync */
+	PSP_DISPLAY_SETBUF_NEXTVSYNC = 1
 };
 
+/** Values for retro compatibility */
+#define PSP_DISPLAY_SETBUF_IMMEDIATE PSP_DISPLAY_SETBUF_NEXTHSYNC
+#define PSP_DISPLAY_SETBUF_NEXTFRAME PSP_DISPLAY_SETBUF_NEXTVSYNC
 
 enum PspDisplayErrorCodes
 {
--- a/src/gu/guInternal.h
+++ b/src/gu/guInternal.h
@@ -645,12 +645,18 @@ static inline void sendCommandf(GECommand cmd, float argument)
 	sendCommandi(cmd, t.i >> 8);
 }
 
-static inline void sendCommandiStall(GECommand cmd, int argument)
-{
-	sendCommandi(cmd, argument);
-
-	if (!gu_object_stack_depth && !gu_curr_context)
-		sceGeListUpdateStallAddr(ge_list_executed[0], gu_list->current);
+static inline int _sceGuUpdateStallAddr(void) {
+	if (gu_curr_context == GU_DIRECT) {
+		// Just if there are no objects in the stack (no guBeginObject)
+		if (!gu_object_stack_depth) {
+			int res;
+			res = sceGeListUpdateStallAddr(ge_list_executed[0], gu_list->current);
+			if (res < 0) {
+				return res;
+			}
+		}
+	} 
+	return 0;
 }
 
 #endif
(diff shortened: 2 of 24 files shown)


src/display/pspdisplay.h
```diff
@@ -31,12 +31,15 @@ enum PspDisplayPixelFormats {
...
-	/** Buffer change effective next frame */
-	PSP_DISPLAY_SETBUF_NEXTFRAME = 1
+	/** Buffer change effective next hsync */
+	PSP_DISPLAY_SETBUF_NEXTHSYNC = 0,
+	/** Buffer change effective next vsync */
+	PSP_DISPLAY_SETBUF_NEXTVSYNC = 1
 };
 
+/** Values for retro compatibility */
+#define PSP_DISPLAY_SETBUF_IMMEDIATE PSP_DISPLAY_SETBUF_NEXTHSYNC
```
fjtrujy, line 41: Keeping previous values for retro compatibility

src/gu/guInternal.h
```diff
@@ -116,7 +116,7 @@ extern GuLightSettings light_settings[4];
 
 void callbackSig(int id, void *arg);
 void callbackFin(int id, void *arg);
-void resetValues();
+void _sceGuResetGlobalVariables();
```
fjtrujy, line 119: use better naming to avoid names conflicts when linking binary

src/gu/sceGuFinishId.c
fjtrujy, line 1: The same content was before in `sceGuFinish`

src/gu/sceGuSignal.c
```diff
@@ -13,11 +13,11 @@ void sceGuSignal(int signal, int argument)
 	sendCommandi(SIGNAL, ((signal & 0xff) << 16) | (argument & 0xffff));
 	sendCommandi(END, 0);
 
-	if (GU_SIGNAL_PAUSE == 3)
+	if (signal == GU_SIGNAL_PAUSE)
```
sharkwouter, line 16: Wait, does this fix a bug? That old code looks weird.

src/gu/sceGuStart.c
```diff
@@ -10,12 +10,16 @@
...
 #include <pspge.h>
+#include <pspuser.h>
 
 void sceGuStart(int ctype, void *list)
 {
+	int intr;
 	GuContext *context = &gu_contexts[ctype];
 	unsigned int *local_list = (unsigned int *)(((unsigned int)list) | 0x40000000);
 
+	intr = sceKernelCpuSuspendIntr();
```
sharkwouter, line 21: Is this expensive to do? Do we have a chance to lose input because of this?

src/gu/guInternal.h
```diff
@@ -645,12 +645,18 @@ static inline void sendCommandf(GECommand cmd, float argument)
 	sendCommandi(cmd, t.i >> 8);
 }
 
-static inline void sendCommandiStall(GECommand cmd, int argument)
-{
-	sendCommandi(cmd, argument);
-
-	if (!gu_object_stack_depth && !gu_curr_context)
-		sceGeListUpdateStallAddr(ge_list_executed[0], gu_list->current);
+static inline int _sceGuUpdateStallAddr(void) {
```
sharkwouter, line 648: So if I understand correctly the stalli function was used to make sure that if you are using direct mode, the function you called before cannot be interrupted? This looks more readable, but I do wonder how this looks in cpu cycles.

sharkwouter (approved):
I'm not able to do testing with this this weekend unfortunately, but the code looks pretty good to me. I just had some small things I don't understand, but I did not spot any issues.

src/gu/guInternal.h
fjtrujy: I think the GCC compiler is clever enough like to put the exactly same ASM code as before

src/gu/sceGuSignal.c
fjtrujy: basically yes.... it was a bug

src/gu/sceGuStart.c
fjtrujy: it shouldn't be expensive, is the best way to assure atomic operations

sharkwouter: Fair

PR #289, opened 2025-05-13
Remove magic numbers from clear function

fjtrujy:


sharkwouter (approved):

sharkwouter:
Thanks! That is better

PR #291, opened 2025-05-13
fix: replace a statement by an assigment

sblondon:
The statement had no effect. It's in a block intended for initialization (as shown in the line 639) so an assigment was probably the intended behaviour.

The graphical output is the same with or without this patch.

Detected by a warning from gcc:

spharm.c: In function ‘SpharmGenTest’:
spharm.c:640:15: warning: statement with no effect [-Wunused-value]
  640 |       objtype == 0;
      |       ~~~~~~~~^~~~


sharkwouter (approved):
Good find, this looks much better.

PR #293, opened 2025-05-16
Fix mipmapping sample

fjtrujy:


sharkwouter (approved):
Looks good, thanks for the fix!

PR #300, opened 2025-06-08
Improve `sceGuLight*` functions

fjtrujy:
## Description
Some improvements around the `sceguLight*` functions:
- Keeps removing magic numbers
- Reduce the static memory used
- Makes more readable each `sceGuLight*` implementation
- Some other minor changes

Cheers


src/gu/sceGuLight.c
```diff
@@ -10,15 +10,21 @@
...
-	GuLightSettings* settings = &light_settings[light];
+	int offset = light * 3;
+	int ctype;
 
-	sendCommandf(settings->xpos,position->x);
-	sendCommandf(settings->ypos,position->y);
-	sendCommandf(settings->zpos,position->z);
+	if (components == GU_POWERED_DIFFUSE) {
+		ctype = GU_DIFFUSE_AND_SPECULAR;
+	} else if (components == GU_DIFFUSE_AND_SPECULAR) {
```
sharkwouter, line 18:
```suggestion
	} else if (components == GU_DIFFUSE_AND_SPECULAR || components == 0) {
```
Otherwise the result will be different if the value given is 0.

src/gu/sceGuLight.c
```diff
@@ -10,15 +10,21 @@
...
-	if (components != 8)
-		kind = (components^6) < 1 ? 1 : 0;
-
-	sendCommandi(settings->type,((type & 0x03) << 8)|kind);
+	// Light position
+	sendCommandf(LIGHT0_X + offset, position->x);
+	sendCommandf(LIGHT0_Y + offset, position->y);
+	sendCommandf(LIGHT0_Z + offset, position->z);
+	// Light type and components
+	sendCommandi(LIGHT_TYPE0 + light, (ctype << 8) | (type));
```
sharkwouter, line 29: I think this should be like this for it to match what we had before:
```suggestion
	sendCommandi(LIGHT_TYPE0 + light, ((type & GU_AMBIENT_AND_DIFFUSE) << 8) | ctype);
```

src/gu/sceGuLight.c
```diff
@@ -10,15 +10,21 @@
...
 {
-	GuLightSettings* settings = &light_settings[light];
+	int offset = light * 3;
+	int ctype;
 
-	sendCommandf(settings->xpos,position->x);
-	sendCommandf(settings->ypos,position->y);
-	sendCommandf(settings->zpos,position->z);
+	if (components == GU_POWERED_DIFFUSE) {
+		ctype = GU_DIFFUSE_AND_SPECULAR;
```
sharkwouter, line 17: GU_DIFFUSE matches 2, which was the default before.
```suggestion
		ctype = GU_DIFFUSE;
```

sharkwouter (changes requested):
There are still some changes in the result of the new code compared to the old code. You can see this in the gu/lights example.

src/gu/sceGuLight.c
sharkwouter: I see, then it is probably best to keep it in case the user gives invalid input. Especially since it was allowed before.

src/gu/sceGuLight.c
fjtrujy: I think that sending 0 should use just `DIFFUSE`

sharkwouter: The original code was `kind = (components^6) < 1 ? 1 : 0;`,  which results in kind being set to 1 if components is set to 0 or 6. So setting it to DIFFUSE instead of DIFFUSE_AND_SPECULAR in the case of 0 would be an API change. Do you think the original behavior is a bug?

fjtrujy: To me it looks a bug of our previous implementation…

PR #308, opened 2025-06-27
Adding 30 FPS VSYNC sample

fjtrujy:
## Description
This PR adds a sample to `gu` directory to force a 30 FPS app by using double vsync

### DIFF
--- a/src/samples/Makefile.am
+++ b/src/samples/Makefile.am
@@ -21,6 +21,7 @@ SAMPLES = \
 	gu/clut \
 	gu/common \
 	gu/copy \
+	gu/doublelist \
 	gu/cube \
 	gu/envmap \
 	gu/integerdrawing \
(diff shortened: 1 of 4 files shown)


sharkwouter:
Could you explain how this code works? I'm a bit confused. What makes the sema do something?

fjtrujy:
> Could you explain how this code works? I'm a bit confused. What makes the sema do something?

With the `sceKernelRegisterSubIntrHandler(PSP_VBLANK_INT, 0, on_vblank, NULL);`, the `on_vblank` function is going to be executed each `PSP_VBLANK_INT` interruption.

To force 30 FPS, we must ensure the VBLANK interruption happens exactly twice, in order to achieve this, I have used 2 semaphores, so before swapping buffers, we need to ensure that both semaphores are being signaled already, and this is what `vsync30` method does.

The reason why I'm creating this example is because there are plenty of games whose FPS are set to 30FPS (as SuperMario 64), so this is just an example of how we could implement this.

fjtrujy:
> Ah okay, that makes sense. Would it be possible to add some additional comments to clarify this?

Added as a comment at the top of the example file

sharkwouter (approved):

Issue #309, opened 2025-07-02
PRX builds using psp-cmake fail with _exit linker error and missing exports.exp


santiago046:
When compiling any PRX project using psp-cmake, the build fails with a missing _exit symbol during linking. In addition, if the .exp file is not manually copied to the build directory, the generation of exports.c fails due to Could not open file exports.exp.

## Steps to Reproduce

1. Clone pspsdk

```
git clone https://github.com/pspdev/pspsdk/
```

2. Create any minimal PRX project (e.g., in pspsdk/src/samples/prx/testprx) with the following CMakeLists.txt:

```
cmake_minimum_required(VERSION 3.11)

project(testprx C)

add_prx_module(${PROJECT_NAME}
    main.c
    exports.exp
)

install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.prx DESTINATION .)
```

3. Run:

```
cd pspsdk/src/samples/prx/testprx
psp-cmake -Bbuild
cmake --build build
```

## Output

```
[ 25%] Generating exports.c from exports.exp for target testprx
Could not open file exports.exp
[ 50%] Building C object CMakeFiles/testprx.dir/main.c.obj
[ 75%] Building C object CMakeFiles/testprx.dir/exports.c.obj
[100%] Linking C executable testprx
/root/pspdev/bin/../lib/gcc/psp/.../ld: warning: cannot find entry symbol module_start; defaulting to 00000000
.../libc.a(...abort.o): undefined reference to `_exit'
collect2: error: ld returned 1 exit status
```

**Note:** copying `exports.exp` to the build directory, or specifing the absolute path to it, fixes `Could not open file exports.exp`.

crozone:
@santiago046 I have the same issue with `_exit`:

> /__w/pspdev/pspdev/build/psptoolchain/build/psptoolchain-allegrex/build/newlib/build-psp/psp/newlib/../../../newlib/libc/stdlib/abort.c:59:(.text+0x10): undefined reference to `_exit'

After bisecting my changes, I found it occurs whenever I use any sort of print function, like:

* `pspDebugScreenPrintf()`
* `fprintf()`
* `printf()`

However, the kernel functions for printing like `pspDebugScreenKprintf()` work just fine. So it's definitely something weird with not using kernel libraries - or something like that, I'm a bit out of my depth on this one.

santiago046:
[...]

PR #310, opened 2025-07-02
CMake: Fix exports.exp not being found in add_prx_module

santiago046:
Related to #309

Fixes the specific case where `add_prx_module()` fails to locate `exports.exp` when passed as a relative path.


sharkwouter (approved):
Thanks, that is a good catch.

PR #312, opened 2025-07-03
Add link to other signal interrupt modes in sceGuSignal description

veka0:


fjtrujy:
I'm not aware of this PR, but not sure if we should link to `uofw`

veka0:
I don't know if including `uofw` link is problematic in some way, that was just the first resource I found that documents additional signal interrupts. I don't feel like I know enough about signals to write the description & macros myself at this point, and it also doesn't feel quite right directly copying it from another project. So yeah, not sure what best to do here.

sharkwouter:
I'm not a fan of this approach either. We should have our own documentation, but I do agree that we shouldn't just copy it from uofw. I might be best to close this one for now and get back to it later.

veka0:
Yeah that's probably the way to go then

PR #313, opened 2025-07-03
Clarify csa parameter in sceGuClutMode documentation

veka0:
This PR clarifies behavior of csa parameter in sceGuClutMode. 

Previously it was described as `Read-out start location (16-palette units)` which implies that it offsets the palette by 16 colors while in reality csa value is shifted to the left by 4 and bitwise ORed with color index. This behavior differs from arithmetic addition for color index values larger than 15 as in that case color index bits will overlap csa bits and create results that are very different from expected 16 unit palette shift.


fjtrujy (approved):
LGTM!

sharkwouter:
Thanks!

PR #314, opened 2025-07-03
Clarify sceGuPixelMask documentation

veka0:
From my testing, `sceGuPixelMask` doesn't support filtering against individual bits, only ignoring writes to entire color channels, similar to its OpenGL counterpart.

In particular, only the 8th bit of each color component seems to be checked by GE, so a value of 128 and above will make GE ignore writes to that color channel.

### DIFF
--- a/src/gu/pspgu.h
+++ b/src/gu/pspgu.h
@@ -955,9 +955,16 @@ void sceGuClearDepth(unsigned int depth);
 void sceGuClearStencil(unsigned int stencil);
 
 /**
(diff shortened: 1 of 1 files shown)


veka0:
Although, this states that unlike modern GPUs, PSP does allow to mask individual bits https://www.ppsspp.org/blog/lens-flare-burnout-dominator

It's weird because I haven't observed this behavior in my tests, but I will do more tests just in case.

PR #315, opened 2025-07-03
Change sceGuViewport params from int to float

veka0:
This enables more flexibility in configuring viewport as you'd be able to specify it with subpixel precision. 

sceGuViewport can still be used as before, with integer parameters and behavior of existing code will remain the same, however if someone for some reason is already using sceGuViewport with float parameters, by relying on conversion from float to int to truncate fractional part, with this change behavior of such implementations will be slightly different as viewport transform may be off by up to 1 pixel. To me it seems unlikely that someone was using it in this way, or that it will have a major impact and break something, but that's not up to me to decide.

### DIFF
--- a/src/gu/pspgu.h
+++ b/src/gu/pspgu.h
@@ -1460,7 +1460,7 @@ void sceGuScissor(int x, int y, int w, int h);
   * @param width - Width of viewport
   * @param height - Height of viewport
 **/
-void sceGuViewport(int cx, int cy, int width, int height);
+void sceGuViewport(float cx, float cy, float width, float height);
 
 /**
   * Draw bezier surface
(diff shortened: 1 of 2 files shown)


fjtrujy (changes requested):
I think this PR is a mistake...
If you want to offer `sceGuViewport` with `float`  values, we can, but please let's keep the `int` one with the original name, as it follow same API as the official `PSPSDK`

so create if you want:
```c
void sceGuViewportF(float cx, float cy, float width, float height)
```

veka0:
The idea behind this PR was to extend already existing functionality in a backwards compatible(-ish) way. But it's fair if you don't want to touch existing foundational functions like this one, I think in that case it might be a good idea to drop this PR & potentially consider adding that function if and when a real use case appears for it in the wild.

sharkwouter:
Since this hasn't been updated for a couple of months I will close this PR. If you wish to continue to work on it, feel free to reopen and update it. We're always happy to get new contributions. Thanks for working to contribute.

PR #316, opened 2025-07-03
Clarify sceGuLogicalOp documentation that color logic op can be used with blending

veka0:
This PR clarifies that blending is allowed to be used with color logic operations. While it is certainly not a common use case, because OpenGL disallows using both at the same time, there can be an assumption that GE functions in the same way while that's not the case.


sharkwouter:
Thanks for the PR!

PR #317, opened 2025-07-04
Add sceJpegDecodeMJpegYCbCr, sceJpegCsc and sceJpegGetOutputInfo

RegularRabbit05:


RegularRabbit05:
Sorry for the wait (and multiple commits) as I got the signature wrong.

With this code i get the following result on ppsspp (with wrong colors apparently so i need to look into that)
image
As for hardware on a psp i get -2140864490 / 0x80650016 from sceJpegDecodeMJpegYCbCr (which should mean unsupported sampling) for most images while some load but look scrambled but all of this could be an error on my end so if anyone can test it too as a double check would be appreciated

RegularRabbit05:
Seems like the functions currently in this header might work specifically with yuvj420p chromas on real hardware vs any jpeg on ppsspp, we'll research more on it and update the header accordingly

sharkwouter (approved):

PR #320, opened 2025-07-13
Creating `libpspgud` library

fjtrujy:
## Description
This PR creates a `libpspgud.a` library. This library is identical to `libpspgu.a`, differing only in the inclusion of numerous assertions and `printf` calls.

This is quite useful for debugging purposes, as it easily reveals if you're doing something not allowed.
It is also it is useful to identify what kind of values are going to be added into the GPU queue.

I'm not sure if some assertions are wrong or some of them are missing, so please, take a look to them.

Cheers.

### DIFF
--- a/src/gu/sceGuTexWrap.c
+++ b/src/gu/sceGuTexWrap.c
@@ -10,5 +10,13 @@
 
 void sceGuTexWrap(int u, int v)
 {
-	sendCommandi(TEX_WRAP, (v << 8) | (u));
+#ifdef GU_DEBUG
+	printf("sceGuTexWrap(%d, %d);\n", u, v);
+	assert(gu_init && "GU not initialized");
+	assert((u == GU_CLAMP || u == GU_REPEAT) && "Invalid U wrap mode");
+	assert((v == GU_CLAMP || v == GU_REPEAT) && "Invalid V wrap mode");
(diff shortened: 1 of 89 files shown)


fjtrujy:
> I can still test this and look into it more deeply, but since there is no code in here that will break existing code, this can be safely merged.

No problem, take your time

src/gu/sceGuTexWrap.c
```diff
@@ -10,5 +10,13 @@
 
 void sceGuTexWrap(int u, int v)
 {
-	sendCommandi(TEX_WRAP, (v << 8) | (u));
+#ifdef GU_DEBUG
+	printf("sceGuTexWrap(%d, %d);\n", u, v);
+	assert(gu_init && "GU not initialized");
+	assert((u == GU_CLAMP || u == GU_REPEAT) && "Invalid U wrap mode");
+	assert((v == GU_CLAMP || v == GU_REPEAT) && "Invalid V wrap mode");
```
sharkwouter, line 17: Could you explain why this is wrong? I have seen this used before iirc.

sharkwouter: I understand this now, never mind.

sharkwouter (approved):

PR #324, opened 2025-08-14
Adding `pixelMask` example

fjtrujy:
## Description
Adding a GU example where we show how to use the `sceGuPixelMask` functionality.

This test uses a texture that is split into 4 different colored columns: RED, GREEN, BLUE, and WHITE. 
Thanks to `sceGuPixelMask`, we can just draw concrete pixels on the screen, ignoring the masked ones. 

Every 2 seconds, we change the mask values, generating:

all
red
green
blue

Cheers


sharkwouter (approved):
Works well, I tested it on a real PSP.

PR #326, opened 2025-08-29
fix mistake in documentation of sceGuTexMode

Mahmoud1205:
sceGuTexMode documentation said that maxmips argument is the number of mips: which is incorrect. maxmips is the index of the maximum mip level used. And the documentation said that the range of this argument is 0-8 , which is incorrect also and the correct range is 0-7.


sharkwouter (approved):
Nice find, thanks! The contribution is really appreciated!

PR #328, opened 2025-09-05
refactor: Update `SceModule` definition

GrayJack:
Update the `SceModule` definition according to RE done on uofw.

According to some talk on discord, this type definition never changed between firmware versions. So this PR also fixes a bug where there should be 160 bits between `stub_size` and `entry_addr` fields, but there was `128` bits between those two fields.

### DIFF
--- a/src/base/psptypes.h
+++ b/src/base/psptypes.h
@@ -427,6 +427,9 @@ typedef struct ScePspDateTime {
 	unsigned int 	microsecond;
 } ScePspDateTime;
 
+/* Thread entry function type. */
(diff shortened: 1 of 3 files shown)


GrayJack:
Oops, there was a silly type error, I updated it

sharkwouter:
Just to let you know, I have not forgotten about this, but I don't know if I'll be able to test this this week. Next week I'll be able to test and merge this, though.

sharkwouter:
That helps! Thanks for the input @JoseAaronLopezGarcia  :)

PR #329, opened 2025-09-28
Fix CMake prx exports build issue for projects building multiple prx's.

meetpatty:
Currently the generated exports.c is output to CMAKE_BINARY_DIR the top level build directory. This causes build issues in projects with multiple prx's as they end up building against an exports.c generated from another prx module. Use CMAKE_CURRENT_BINARY_DIR instead.


fjtrujy (approved):
LGTM!

fjtrujy:
Please @sharkwouter review and merge it if you agree.

Cheers!

sharkwouter (approved):

PR #358, opened 2026-03-17
Fix the debug keyboard failing to type.

SilverNexus2:
A bug introduced in 332dd434f1ce090652a94011a434fe61774cb61f caused the keyboard to locally modify the source strings, and then discard the changes without writing them back to the original string.
This change drops the use of snprintf to perform the changes entirely, opting instead for twiddling the characters in-place. This required changing the bounds check to ensure there remains room for the null-terminator, and a call to strlen was refactored to be stored for the local scope.


diamant3 (approved):

sharkwouter:
Thanks for contributing :)

PR #360, opened 2026-04-17
fix pspSdkInstallNoPlainModuleCheckPatch

JoseAaronLopezGarcia:
It seems that in older versions of the SDK, sceKernelProbeExecutableObject and sceKernelCheckPspConfig were function pointers but they are now stubs, causing pspSdkInstallNoPlainModuleCheckPatch to crash.
This PR fixes it.


fjtrujy (approved):
LGTM!
@sharkwouter ??

PR #368, opened 2026-06-14
feat(tools, base): Add support to generate `PBOOT.PBP` and to sign VITA PSPemu `PBOOT.PBP`

GrayJack:
- `PrxEncrypter` updated to support signing VITA PSPemu `PBOOT.PBP`
- `mksfoex`updated with argument to generate `PBOOT.PBP`
- Add support for `ENCRYPT=2` in `build.mak`
- Add support for `ENCRYPT` in `build_prx.mak`
- Add support to generate `PBOOT.PBP` and sign VITA PSPemu `PBOOT.PBP` on CMAKE `create_pbp_file`


JoseAaronLopezGarcia (approved):
These changes are very important, specially to preserve the ability to generate Vita OFW compatible PSP Homebrew and also forms the basis for ARK's loader on PS Vita.

I have personally tested and confirmed the changes work for generating signed/encrypted PBOOT.PBP that the PS Vita accepts as valid and do not break any previous builds such as PSP OFW encrypted binaries.

GrayJack:
This last commit is a fix I did locally so many months ago, and now that I did a new environment, I got it again

```
CMake Error at .CMakeLists.txt:3 (project):
  The CMAKE_C_COMPILER:

    psp-gcc

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
```

Passing a full path solves the issue and don't break when the binaries are in the path

The change is small enough that I decided to add in this PR

src/base/build.mak
```diff
@@ -124,9 +124,28 @@ ifndef PSP_EBOOT_PSAR
 PSP_EBOOT_PSAR = NULL
 endif
 
+ifndef PSP_PRODUCE_PBOOT
+ifeq ($(ENCRYPT), 2)
+PSP_PRODUCE_PBOOT = 1
+
```
sharkwouter:
```suggestion
```

sharkwouter (commented):
This looks fine, but I'll still need some time to test this. I did have one small comment.

sharkwouter (approved):

PR #371, opened 2026-06-26
Add guGetInit function

sharkwouter:
This function is needed to make sure we can always spawn the keyboard or notifications while SDL is being used for example, but I'm sure there are other use cases. The variable that is exposed here was already used internally a lot.


sajattack:
if it's not sce it should be pspGuGetInit no?

sajattack (approved):

PR #381, opened 2026-07-25
feat(kernel): Add stubs for `sceAmctrl_driver`

GrayJack:
We already have the prototypes (added on #348), but we lack the stubs, so here they are


JoseAaronLopezGarcia (approved):
LGTM

PR #383, opened 2026-08-27
feat: Make all stubs weak imports

GrayJack:
Make all stubs weak. This makes modules not fail to load when a linked lib was not loaded yet.

This is benefitial mostly for plugins, as they are often loaded before all system modules were started. And CFW have API that allows to inject a module to be loaded as early as the second module the system loads; in this edge case, all modules but sysmem libraries ideally should be weak.

I also noticed that some stubs were with the version part as `0011` which is the system module version used on `FW>=6.60` on some modules, this will cause any module linking to them to fail loading on firmware version `<6.60` as the system checks the stubs version on load if the library module was already loaded (if the imports are below or equal to the loaded library version, the module fails to load). On the cases of library modules allowed to be loaded multiple instaces of the module, at least one loaded version must match the imported stub version. I did not test the case of a weak imported with version bigger than the version that will be loaded after, but my best guess is that some error will occur in this case as well. To avoid this, I put then all to zero on the SDK stubs.

### DIFF
--- a/src/audio/sceAudio_driver.S
+++ b/src/audio/sceAudio_driver.S
@@ -3,7 +3,7 @@
 #include "pspimport.s"
 
 #ifdef F_sceAudio_driver_0000
-	IMPORT_START	"sceAudio_driver",0x00010000
+	IMPORT_START	"sceAudio_driver",0x00090000
 #endif
 
 #ifdef F_sceAudio_driver_0001
--- a/src/mpeg/sceMpegbase_driver.S
+++ b/src/mpeg/sceMpegbase_driver.S
@@ -3,7 +3,7 @@
 #include "pspimport.s"
 
 #ifdef F_sceMpegbase_driver_0000
-	IMPORT_START	"sceMpegbase_driver",0x00010000
+	IMPORT_START	"sceMpegbase_driver",0x00090000
 #endif
 #ifdef F_sceMpegbase_driver_0001
 	IMPORT_FUNC	"sceMpegbase_driver",0xBE45C284,sceMpegBaseYCrCbCopyVme
--- a/src/sascore/sceSasCore.S
+++ b/src/sascore/sceSasCore.S
@@ -3,7 +3,7 @@
 #include "pspimport.s"
 
 #ifdef F_sceSasCore_0000
-	IMPORT_START	"sceSasCore",0x40090011
+	IMPORT_START	"sceSasCore",0x40090000
 #endif
 #ifdef F___sceSasInit_0001
 	IMPORT_FUNC	"sceSasCore",0x42778A9F,__sceSasInit
--- a/src/audio/sceAudio.S
+++ b/src/audio/sceAudio.S
@@ -3,7 +3,7 @@
 #include "pspimport.s"
 
 #ifdef F_sceAudio_0000
-	IMPORT_START	"sceAudio",0x40010000
+	IMPORT_START	"sceAudio",0x40090000
 #endif
 #ifdef F_sceAudio_0001
 	IMPORT_FUNC	"sceAudio",0x8C1009B2,sceAudioOutput
--- a/src/ctrl/sceCtrl.S
+++ b/src/ctrl/sceCtrl.S
@@ -3,7 +3,7 @@
 #include "pspimport.s"
 
 #ifdef F_sceCtrl_0000
-	IMPORT_START	"sceCtrl",0x40010000
+	IMPORT_START	"sceCtrl",0x40090000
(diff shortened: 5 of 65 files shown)


src/audio/sceAudio_driver.S
```diff
@@ -3,114 +3,114 @@
...
-	IMPORT_START	"sceAudio_driver",0x00010000
+	IMPORT_START	"sceAudio_driver",0x00090000
 #endif
 
 #ifdef F_sceAudio_driver_0001
 	IMPORT_FUNC	"sceAudio_driver",0x80F1F7E0,sceAudioInit
 #endif
 
-#ifdef F_sceAudio_driver_0002 
+#ifdef F_sceAudio_driver_0002
```
rofl0r: these gratuitous whitespace changes clutter the diff

src/mpeg/sceMpegbase_driver.S
```diff
@@ -17,21 +17,21 @@
 #ifdef F_sceMpegbase_driver_0004
 	IMPORT_FUNC "sceMpegbase_driver",0x0530BE4E,sceMpegbase_driver_0530BE4E
 #endif
-#ifdef F_sceMpegbase_driver_0005
+#ifdef F_sceMpegbase_driver_0005
```
rofl0r: i'm staring at these diff lines since a minute but cannot see an actual difference. did you change the line ending ?

src/audio/sceAudio_driver.S
GrayJack: Sure, it was automatic from editor, I didn't even noticed, but considering the majority of stub files don't have a end of line whitespace, removing then just help normalize the stub file format

rofl0r: maybe, but it should be a separate cleanup PR, not smuggled in a unrelated change. everyone has different editor settings, so if this kind of change wouldnt be pointed out, 90% of all PRs would just be whitespace noise.

src/mpeg/sceMpegbase_driver.S
GrayJack: I did not change line ending, and would not make sense as it only happens on some lines.

I did a local diff and this are not even showing up in this file, only the one in line 6, maybe it is a github algorithm thing

src/audio/sceAudio_driver.S
GrayJack: I disagree that it would need a separete PR, but indeed not ideal on one commit.

I also disagree on your second affirmation, as the solution is somewhat simple, decide on a project code style that is easily automated with a formatter, that way it can be auto-checked on CI and also easily fixable by PR authors while also not being a waste of time for reviewers. So making maintaining easier, no whitespace noise on PR, no waste time on review that end up being style change (as it will be none as the CI will fail).

GrayJack: Let me fix the whitespace for this PR in a few minutes.

src/mpeg/sceMpegbase_driver.S
rofl0r: last time it happened, it was in a file with mixed line endings. and it was your editor - again - that messed it up: https://github.com/pspdev/pspsdk/pull/342#discussion_r2659293196

src/sascore/sceSasCore.S
```diff
@@ -3,7 +3,7 @@
 #include "pspimport.s"
 
 #ifdef F_sceSasCore_0000
-	IMPORT_START	"sceSasCore",0x40090011
+	IMPORT_START	"sceSasCore",0x40090000
```
rofl0r, line 6: this seems to use the same address as sceSircs further down, and sceUsb...
edit: hmm. seems to be a common pattern here, even in the previous addresses? not familiar what these actually represent so disregard.

src/audio/sceAudio_driver.S
rofl0r: thanks for the cleanups; now it's much easier to see what the commit changes.

src/sascore/sceSasCore.S
GrayJack: The first half is library import flags, the second half is version.

I bit-or'ed 0x8 (weak import flag) on the first half to make the stub weak.

The version I put them to zero, as I explained on the PR main text.

rofl0r (commented):
since you requested a review: optically looks good now, but i cant comment on the change of the constants, as i dont really have an insight what the effect of changing them might cause.

JoseAaronLopezGarcia (approved):
This is a necessary change, specially for low level kernel development.

PR #386, opened 2026-09-14
Improve headers consistency

Ryp:
Hello!

I ran this cleanup pass so that the external python parser that I use doesn't choke on some of the unusual patterns present in the headers. The build passes, I only did simple modifications.

See also: https://github.com/zPSP-Dev/Zig-PSP/blob/trunk/tools/generate_bindings.py#L101

Thanks


GrayJack:
Do we have the stubs for sceImpose?

Ryp:
For usermode, here: https://github.com/pspdev/pspsdk/blob/master/src/user/sceImpose.S

GrayJack (approved):
LGTM

rofl0r:
this looks like gratuitous changes that could break stuff for questionable gain.

bucanero:
> this looks like gratuitous changes that could break stuff for questionable gain.

I agree here. This change could impact on homebrew apps downstream , and it seems mostly cosmetic. If you want to define new types it's okay, but current methods should remain unchanged.

GrayJack:
That is a fair concern.

As far as I know from C standard (and for what  I know about C++), as long as the type name remains together with the typedef alias, both `⟨type⟩ TypeName` and `TypeAliasName` works interchangeably by both languages (with the exeception for self field type definition, i.e. inside the type definition itself you can't use `TypeAliasName` for the type of fields).

I don't know about non-standard C, but they usually only expand what is allowed.

Basically, this

```c
struct MyStruct { /**/ };
```

To

```c
typedef struct MyStruct { /**/ } MyStructAlias;
```

Using `struct MyStruct` or `MyStructAlias` is interchangeable (outside the `MyStruct` fields) in C and C++. No breaking changes even changing the function prototypes.

But this

```c
struct MyStruct { /**/ };
```

To

```c
typedef struct { /**/ } MyStructAlias;
```

Is a breaking change, as `struct MyStruct` will no longer be allowed in C, and I heard of weird issues on C++ depending on the standard used (or something like that).

Ryp:
I get the feedback, it's not adding much except for the `pspimpose.h` header which was just missing. To the best of my knowledge this has a small blast radius and arguably iteratively improves the SDK

Ryp:
I tested on real HW in user mode, with `sceImposeGetBacklightOffTime` and `sceImposeSetBacklightOffTime` with the NIDs provided and they work.

PR #390, opened 2026-09-20
Import known NID names from PSPLibDoc

Ryp:
Small header changes as well, but I doubt anyone is using a `sceXXX_<NID>` directly


sharkwouter (approved):

Ryp:
Just added a minor commit for maximum C compat. C is weird.

